PR #20, opened 2017-11-13
use dotenv to seperate config data from code.

strand:


strand:
Before this PR is accepted, please ensure that the README has been updated with instructions on getting your .env populated, along with providing an example .env file.

strand:
PR replaced by #21.

PR #22, opened 2017-11-15
Loads appropriate text for events in timeline

RedSquirrelious:
loads event text from event type, event.data.displayText, or stringifies event.data


src/components/Timeline/Events.js
```diff
@@ -24,10 +24,13 @@ export const Events = (events, eventActions, eventTypeActions, ticketId, inciden
...
 
-const LoadTextFromEventType = (eventTypeId, eventTypes) => {
-  const eventType = eventTypes[eventTypeId]
-  if(!eventType) return 'Loading event information...'
-  return eventType.name
+const LoadTextFromEvent = (event, eventTypes) => {
+  const eventType = eventTypes[event.eventTypeId]
+  if (eventType) 
+    return eventType.displayTemplate.pattern ? eventType.displayTemplate.pattern 
+    : (eventType.name ? eventType.name : 'Loading event information...')
```
maggiepint: this can be expressed as `eventType.name || 'Loading event information...'`

src/components/Timeline/Events.js
```diff
@@ -24,10 +24,13 @@ export const Events = (events, eventActions, eventTypeActions, ticketId, inciden
...
+  else
+    return event.data.displayText ? event.data.displayText : JSON.stringify(event.data)
```
maggiepint: same here, `event.data.displayText || JSON.stringify(event.data)`

src/components/Timeline/Events.js
```diff
@@ -24,10 +24,21 @@ export const Events = (events, eventActions, eventTypeActions, ticketId, inciden
...
 }
 
-const LoadTextFromEventType = (eventTypeId, eventTypes) => {
-  const eventType = eventTypes[eventTypeId]
-  if(!eventType) return 'Loading event information...'
-  return eventType.name
+const LoadTextFromEvent = (event, eventTypes) => {
+  const eventType = eventTypes[event.eventTypeId]
+  if(!eventType) 
+    if (!event.data.displayText)
```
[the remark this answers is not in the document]
RedSquirrelious: I thought that we were going with Playbook Event Types first.

src/components/Timeline/Events.js
```diff
@@ -24,10 +24,31 @@ export const Events = (events, eventActions, eventTypeActions, ticketId, inciden
...
+const HasValidDisplayTemplatePattern = (eventType) => {
+  return eventType.displayTemplate && eventType.displayTemplate.pattern && eventType.displayTemplate.pattern.length > 0 ? true : false
+}
+
+const HasValidName = (eventType) => {
+  return eventType.name && eventType.name.length > 0 ? true: false
+}
+
+const HasValidData = (event) => {
+  return event.data ? true: false
```
pdimitratos: One way to simplify this statement is:

return !!(event.data)

src/components/Timeline/Events.js
```diff
@@ -24,10 +24,31 @@ export const Events = (events, eventActions, eventTypeActions, ticketId, inciden
...
+const LoadTextFromEvent = (event, eventTypes) => {
+  const eventType = eventTypes[event.eventTypeId]
+  if (eventType)
+    return  HasValidDisplayTemplatePattern(eventType)? eventType.displayTemplate.pattern
+    : (HasValidName(eventType) ? eventType.name : 'Loading event information...')
+  else if (HasValidData(event))
+    return HasValidDisplayText(event.data) ? event.data.displayText : JSON.stringify(event.data)
+  else
+    return 'Loading event information...'
+}
```
pdimitratos: If the order of precedence we want is:

1. Display Template
2. Display Text
3. Event Type Name
4. Event Data Json

Then this conditional can't branch on having an eventType, since a valid event type with an invalid template would use its name rather than the event's display text. Rather than checking for eventType prior to calling HasValidDisplayTemplatePattern, the validation of the eventType object existing could be moved into that function (which would just add 'eventType &&' between 'return' and 'eventType.displayTemplate'). If all the validation happens in the HasValid functions, LoadTextFromEvent becomes simpler, :

return HasValidDisplayTemplatePattern(eventType)? eventType.displayTemplate.pattern 
: HasValidDisplayText(event) ? event.data.displayText
: HasValidName(eventType) ? eventType.name 
: HasValidData(event) ? JSON.stringify(event.data)
: 'This event has no text'

Now that we have a reducer that tracks which event types are being loaded, information from that reducer can be used to determine whether we should display 'Loading event information...', but that doesn't need to be part of this pull request.

RedSquirrelious: @pdimitratos @maggiepint  code updated per your suggestions.  Order is displayTemplate, displayText, event.name, event.data, 'no' data message.  Loader method is simplified.

PR #24, opened 2017-11-17
Features/playbook event text integration

pdimitratos:
Display text templates now populate from sources. Extracted a lot of Playbook business logic to a new playbookService.


src/components/Timeline/Event.js
```diff
@@ -1,11 +1,13 @@
...
+import { connect } from 'react-redux'
 import PropTypes from 'prop-types'
 import moment from 'moment'
 import { Card, CardHeader, CardText } from 'material-ui/Card'
 import BootstrapPlaybook from './Playbook/BootstrapPlaybook'
 import Playbook from './Playbook/Playbook'
+import { LoadTextFromEvent } from '../../services/playbookService'
 
-const Event = ({ text, time, backgroundColor, incidentId, ticketId, eventTypeId, eventId, eventActions, eventTypeActions }) => {
+export const Event = ({ text, time, backgroundColor, incidentId, ticketId, eventType, eventTypeId, eventId, eventActions, eventTypeActions }) => {
```
strand: Can we break up this line? What line length conventions should we have for this project? (I can live with 110, but prefer to live under 80-chars :P )

src/components/Timeline/Playbook/Playbook.js
```diff
@@ -1,9 +1,14 @@
 import React from 'react'
 import { connect } from 'react-redux'
-import moment from 'moment'
-import ByPath from 'object-path'
 import Play from './Play'
-import { selectSourceObject } from './Play'
+import {
+    selectSourceObject,
+    GetComparisonValue,
```
strand: Why do we need these functions other than `selectSourceObject` imported?

src/services/playbookService.js
```diff
@@ -0,0 +1,138 @@
...
+    switch(sourceObjectEnum){
+        case 0: return event
+        case 1: return ticket
+        case 2: return eventType
+        case 3: return engagement
+        default: return null //not sure if there's a better behavior for undefined enum
+    }
+}
+
+export const GetComparisonValue = (condition) => {
```
strand: This function and the ones below appear to be test helper functions, do they need to exist in the `playbookService.js`?

src/components/Timeline/Playbook/Playbook.js
```diff
@@ -60,79 +65,5 @@ export const mapStateToProps = (state, ownProps) => {
     }
 }
 
-export default connect(mapStateToProps)(Playbook)
```
strand: Oh! I see, we're extracting them up a level.

src/components/Timeline/Event.js
pdimitratos: We don't really have any defined style conventions (aside from not using semicolons in JS). I wouldn't mind adding a linter rule to keep things within a certain line length (warning at first until we get everything already existing updated, then upgrading to error to prevent regression). Changing this to align arguments vertically.

strand:
In general this is looking good, though I think we should just import the test functions in our test environ.

src/components/Timeline/Playbook/Playbook.js
pdimitratos: We don't; removed.

src/services/playbookService.js
pdimitratos: These are business logic functions. The 'testable' versions allow injection of function dependencies so that each can be tested in isolation; the 'testable' version are only meant to be used by testing, the non-testable versions are the actual implementations with the real dependencies pre-injected.

src/components/Timeline/Event.js
```diff
@@ -60,19 +60,20 @@ Event.propTypes = {
...
     const ticket = state.tickets.map[ownProps.ticketId]
     const engagement = state.engagements.list.find(engagement => engagement.id === ownProps.engagementId)
     return {
         ...ownProps,
         ticket,
         engagement,
         eventId: event.id,
-        time: moment(event.occurred),
+        eventTypeId: event.eventTypeId,
+        time: moment(event.occurred ? event.occurred: event.Occurred),
```
strand: What is the difference between `event.occurred` and `event.Occurred`?

pdimitratos: JSON serialization inconsistencies between SignalR and the REST API on the Gateway. Once that's fixed in the Gateway, we should revert this line.

PR #26, opened 2017-11-23
default sort order is descending

RedSquirrelious:
Added optional third argument to 'paginated' that defined defaultSortOrder.  Removed method called later that somehow set the order to ascending.

### DIFF
--- a/src/components/Timeline/Timeline.js
+++ b/src/components/Timeline/Timeline.js
@@ -32,7 +32,6 @@ class Timeline extends Component {
 
 const updatePagination = (dispatch, incidentIds) => {
     dispatch(eventActions.pagination.filter(incidentIds[0].toString()))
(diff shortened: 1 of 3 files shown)


src/components/Timeline/Timeline.js
```diff
@@ -32,7 +32,7 @@ class Timeline extends Component {
 
 const updatePagination = (dispatch, incidentIds) => {
     dispatch(eventActions.pagination.filter(incidentIds[0].toString()))
-    dispatch(eventActions.pagination.sort('occurred'))
+    // dispatch(eventActions.pagination.sort('occurred'))
```
strand: Delete instead of commenting out.

RedSquirrelious: Deleted!  Thank you!

PR #28, opened 2017-11-29
Features/engagements add remove participants

RedSquirrelious:


src/components/Search/CreateIncident.js
```diff
@@ -3,23 +3,31 @@ import { connect } from 'react-redux'
...
-const onSubmit = (input, ticketSystem, history) => () => {
 
-    history.push(/tickets/ + input)
+const onSubmit = (input, history) => () => {
+    if (input)
+    {history.push(/tickets/ + input)}
 }
 
-export const CreateIncident = ({input, ticketSystem, creationError, history, dispatch}) => {
+const submitIfEnter = (input, history) => (event) => {
```
maggiepint: This should actually be done by way of a form post. If we get the HTML inputs into a form, then the enter key will automatically post the form. That's the idiomatic way to do something like this.

PR #30, opened 2017-12-02
Replaced global access of configuration with importing config module.

macrobug:
This helps with testing, decoupling, and makes it easier to enable scenarios such as converting into an installable component.

### DIFF
--- a/cfg/dist.js
+++ b/cfg/dist.js
@@ -19,16 +19,7 @@ let config = Object.assign({}, baseConfig, {
     new webpack.optimize.DedupePlugin(),
     new webpack.DefinePlugin({
       'process.env.NODE_ENV': '"production"',
-      'BASE_URL': constants.baseUrl,
-      'AUTH_REDIRECT_URI': constants.authRedirectUri,
-      'RETRIES': constants.retries,
-      'RETRY_EXPONENTIAL_BACKOFF_FACTOR': constants.retryExponentialBackoffFactor,
-      'RETRY_MIN_TIMEOUT': constants.retryMinTimeoutInMiliseconds,
-      'RETRY_MAX_TIMEOUT': constants.retryMaxTimeoutInMiliseconds,
-      'TICKET_REFRESH_INTERVAL': constants.ticketRefreshIntervalInSeconds,
-      'AAD_INSTANCE': constants.aadInstance,
-      'AAD_TENANT': constants.aadTenant,
-      'CLIENT_ID': constants.clientId
+      constants
     }),
     new BowerWebpackPlugin({
       searchResolveModulesDirectories: false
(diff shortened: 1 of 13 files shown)


cfg/dist.js
```diff
@@ -26,6 +26,7 @@ let config = Object.assign({}, baseConfig, {
       'RETRY_MIN_TIMEOUT': constants.retryMinTimeoutInMiliseconds,
       'RETRY_MAX_TIMEOUT': constants.retryMaxTimeoutInMiliseconds,
       'TICKET_REFRESH_INTERVAL': constants.ticketRefreshIntervalInSeconds,
+      'AAD_INSTANCE': constants.aadInstance,
```
strand: Can we extract out the duplicate attributes in this and the other `cfg/*.js` files into a shared object so the structure of this object does not get out of sync?

src/config/README.md
```diff
@@ -1,9 +1,13 @@
 # About this folder
 
 This folder holds configuration files for different environments.
+
 You can use it to provide your app with different settings based on the 
 current environment, e.g. to configure different API base urls depending on 
 whether your setup runs in dev mode or is built for distribution.
+
+***Note that there is a Webpack alias defined in /cfg/base.js which redirects `'config'` to `'${projectRoot}/config/$env'`***
```
strand: 🤔 Can you specify _why_ the redirection is interesting or surprising? What in the development process made you decide to update the `README.md`?

macrobug: Good question. The redirection itself is neither interesting or surprising. What was not immediately clear to me is where the redirection alias was defined (note that there are two places for configuration ./cfg/ and ./src/config/).

Also, note that this is not the main README file, but a config specific README file. So, it made sense to me to put the note here to help people understand the config pipeline a bit more.

cfg/dist.js
macrobug: Looking into it...

cfg/test.js
```diff
@@ -45,17 +45,7 @@ var config = {
...
-      'AUTH_REDIRECT_URI': constants.authRedirectUri,
-      'RETRIES': constants.retries,
-      'RETRY_EXPONENTIAL_BACKOFF_FACTOR': constants.retryExponentialBackoffFactor,
-      'RETRY_MIN_TIMEOUT': constants.retryMinTimeoutInMiliseconds,
-      'RETRY_MAX_TIMEOUT': constants.retryMaxTimeoutInMiliseconds,
-      'TICKET_REFRESH_INTERVAL': constants.ticketRefreshIntervalInSeconds,
-      'AAD_TENANT': constants.aadTenant,
-      'CLIENT_ID': constants.clientId
-    })
+    new webpack.DefinePlugin({constants})
```
strand: 👏

PR #38, opened 2017-12-13
Bug 43285: Have ticket ID direct to ICM

macrobug:
Added the ticket systems to the config.
Also, refactored the constants to avoid having both single and double quotes.

### DIFF
--- a/cfg/defaultConstants.js
+++ b/cfg/defaultConstants.js
@@ -1,27 +1,41 @@
 module.exports = {
 
-    baseUrl: JSON.stringify("http://localhost:50000/"),
+  baseUrl: 'http://localhost:50000/',
 
-    authRedirectUri: JSON.stringify("http://localhost:3000/"),
+  authRedirectUri: 'http://localhost:3000/',
 
-    retries: 2,
+  retries: 2,
 
-    retryExponentialBackoffFactor: 2,
+  retryExponentialBackoffFactor: 2,
 
-    retryMinTimeoutInMiliseconds: 50,
+  retryMinTimeoutInMiliseconds: 50,
 
-    retryMaxTimeoutInMiliseconds: 1500,
+  retryMaxTimeoutInMiliseconds: 1500,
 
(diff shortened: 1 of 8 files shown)


cfg/defaultConstants.js
```diff
@@ -1,27 +1,41 @@
...
 
-    aadInstance: "'https://login.microsoftonline.com/'", //Replace only if using AAD non-public instances. Yes, use both single and double quotes.
+  aadTenant: 'Your Azure Active Directory tenant here',
 
-    aadTenant: "'Your Azure Active Directory tenant here'", //Yes, use both single and double quotes.
-
-    clientId: "'Your EventUI Azure Active Directory app registration client ID here'" //Yes, use both single and double quotes.
+  clientId: 'Your EventUI Azure Active Directory app registration client ID here',
 
+  ticketSystems: {
```
maggiepint: Couldn't you use a template literal since this is a JS file?

pdimitratos: The PowerShell configuration builder script needs to be updated to be consistent with the double quotes changes and allow for generating a ticketing system from environment variables before we merge this.

macrobug: I considered doing that, but since we don't have the ticketId variable in that scope this would require to eval() it at run time, or do some sort of string replacement, or a similar approach. 

Another approach would be to convert it into (ticketId) => `http://example1/ticket/${ticketId}/something`, but this will be serialized as JSON anyways, so it'll require some deserialization on the other end. 

At the end, having prefix/suffix instead seemed liked a good simple compromise.

macrobug: on it...

maggiepint: FWIW - JavaScript people love them some Evals - so it's not a weird thing to do at all in the language. That said, I'm okay with your choice.

macrobug: Done!

pdimitratos: I updated the build with the new environment variables, so we're ready to merge.

PR #40, opened 2017-12-18
Highlight animation fix

macrobug:


src/components/Timeline/Event.js
```diff
@@ -31,7 +36,7 @@ export const Event = ({
         >
             <CardHeader
                 title={ticketId ? `${ticketId}: ${text}` : text}
-                subtitle={time ? time.format('LT') : 'Time unknown!'}
+                subtitle={time ? time.format('LTS') : 'Time unknown!'}
```
strand: I wish that LTS were human readable, but I was able to find docs on it for moment.js, so I believe this format string is grokkable.

maggiepint: it stands for Local Time with Seconds

PR #41, opened 2017-12-20
Maintenance/requests in flight

pdimitratos:


test/services/PlaybookServiceTest.js
```diff
@@ -3,239 +3,265 @@ import { expect } from 'chai'
 import moment from 'moment'
 import * as playbook from '../../src/services/playbookService'
 
-describe('TestCondition', function () {
```
pdimitratos: This large diff is because I wrapped all the existing describe statements in a single describe statement to make it more obvious that test failures are related to PlaybookServiceTest.

PR #43, opened 2017-12-21
Removed popup reducer and actions; replaced popup AddEventDialog with…

pdimitratos:
… AddEventCard for an inline form


maggiepint:
Great! Glad to reduce the complexity among everything else.

PR #44, opened 2017-12-21
url filters pull from state - in progress, need assistance

RedSquirrelious:
Right now using componentDidUpdate in Timeline.js, trying to trigger action in getEventTypesActionSet.  @maggiepint


RedSquirrelious:
Made some more changes.  Now slightly confused about how eventTypes are moving around.  They're there when you need them, but I lost track.  I need to look again.  @maggiepint

PR #46, opened 2017-12-27
Features/filters pull from state -- filters pulled out

RedSquirrelious:
Filters transform data.  Including filterActions file shows that and simplifies eventActions.  @pdimitratos


src/actions/eventTypeActions.js
```diff
@@ -30,3 +35,23 @@ export const getEventTypeActionSet = (eventTypeId) => ({
         failureReason
     })
 })
+
+export const getEventTypesActionSet = () => ({
+    try: () => ({
+        type: TRY_GET_EVENT_TYPES
+    }),
+
+    succeed: (eventTypes) => (dispatch) => {
```
pdimitratos: Don't need the extra layer of dispatch anymore, this can just be

(eventTypes) => ({
type: GET_EVENT_TYPES_SUCCESS,
eventTypes
})

src/actions/filterActions.js
```diff
@@ -0,0 +1,122 @@
...
+
+export const changeEventFilter = (history) => (filter) => {
+    getUrlFromFilter(history, filter)
+    return {
+        type: CHANGE_EVENT_FILTER,
+        filter
+    }
+}
+
+export const addFilter = (history) => (filter, eventType) => (dispatch) => {
```
pdimitratos: Since we're not dispatching multiple actions from this, it can lose the extra dispatch call and instead of passing the result of applyFilter to dispatch, just return applyFilter(history)(oldFilter, newFilter)

src/actions/eventTypeActions.js
RedSquirrelious: Fixed.

src/actions/eventTypeActions.js
```diff
@@ -4,13 +4,18 @@ export const TRY_GET_EVENT_TYPE = 'TRY_GET_EVENT_TYPE'
...
+export const GET_EVENT_TYPES_FAILURE = 'GET_EVENT_TYPES_FAILURE'
 
 export const fetchEventType = (eventTypeId) => reduxBackedPromise(
     ['eventTypes/' + eventTypeId],
     getEventTypeActionSet(eventTypeId)
 )
 
-
+export const fetchEventTypes = (history) => reduxBackedPromise(
+    ['eventTypes/'], getEventTypesActionSet(history)
```
pdimitratos: No longer needs history; getEventTypesActionSet doesn't need to be a function anymore.

RedSquirrelious: Fixed

src/actions/filterActions.js
RedSquirrelious: Fixed

src/actions/filterActions.js
```diff
@@ -0,0 +1,122 @@
...
+    if (!newFilter.incidentId) {
+        throw 'Need to filter on incidentId!'
+    }
+    if (!deepEquals(oldFilter, newFilter)) {
+        dispatch(changeEventFilter(history)(newFilter))
+        dispatch(eventActions.fetchEvents(newFilter))
+    }
+}
+
+export const synchronizeFilters = (filter, incidentId, ticketId, history, dispatch) => {
```
pdimitratos: Since this is an action, it should be called like:

dispatch(filterActions.synchronizeFilters(myFilter, myIncidentId, myTicketId, myHistory)

rather than

filterActions.synchronizeFilters(myFilter, myIncidentId, myTicketId, myHistory, dispatch)

the signature can change to 
export const synchronizeFilters = (filter, incidentId, ticketId, history) => {

then just return applyFilter(history)(filter, newFilter) instead of dispatching it.

src/actions/filterActions.js
```diff
@@ -0,0 +1,122 @@
...
+    const newFilter = Object.assign({ incidentId: incidentId, ticketId: ticketId }, filter)
+    dispatch(applyFilter(history)(filter, newFilter))
+}
+
+export const serializeFiltersForUrl = (filters) => {
```
pdimitratos: Anything that doesn't work directly with dispatch and isn't used by something that works directly with dispatch should probably be moved out to a filterService at some point, though that doesn't need to be done as part of this pull request.

src/actions/eventTypeActions.js
RedSquirrelious: Fixed.

src/actions/filterActions.js
RedSquirrelious: Fixed

src/actions/filterActions.js
RedSquirrelious: Makes sense.  Added to backlog.

PR #47, opened 2017-12-27
Maintenance/error handling

pdimitratos:
Display error messages and prevent stop repeated fetch of errored records.


test/reducers/eventTypeReducerTest.js
```diff
@@ -139,47 +139,4 @@ describe('EventType Reducer', function EventTypeReducerTest() {
             })
         })
     })
-    describe('Fetching', function () {
```
pdimitratos: This was redundant to the test of buildFetching

PR #48, opened 2017-12-28
Fixed syntax errors, moved store connection closer to where results a…

pdimitratos:
…re used in DisplayPlaybook


src/components/Timeline/Playbook/DisplayPlaybook.js
```diff
@@ -10,17 +12,59 @@ export const DisplayPlaybook = ({
...
+    </div>
+}
+
+
+export const mapStateToDisplayPlaybookProps = (state, ownProps) => {
+    const auth = state.auth
+
+    const eventType = state.eventTypes.records[ownProps.eventTypeId]
+    const event = Object.values(state.events.list)
+        .find(event => event.id == ownProps.eventId)
```
macrobug: Should we be using === instead of == to avoid unexpected stuff like '' == 0 being true?

maggiepint:
Ha! I had fixed this one on my local and was going to PR it this morning. You beat me to it!

PR #37, opened 2017-12-12
adds filter chips from autocomplete text field or URL, filters update state & URL, filters create Gateway URL

RedSquirrelious:

PR #49, opened 2017-12-28
Filter Events by EventType

RedSquirrelious:
Squashed the changes made in PR #37 up to 42957e1861f49763c33967ea9ffd5cb265b0b9c0


test/actions/FilterActionsTest.js
```diff
@@ -0,0 +1,180 @@
...
+
+    describe('serializeEventTypesForQuery', function() {
+        const eventTypes = [1, 2]
+        const noEventTypes = []
+
+        const expectedResult = 'eventTypes=1&eventTypes=2'
+
+        const result = filterActions.serializeEventTypesForQuery(eventTypes)
+
+        it('should return a serialized array', function() {
```
strand: The expected return value is a string of query parameters, and query params are a method of serializing dictionaries/hashes. Let's specify 'string of query params'. BTW, I dove down a research hole on query params and discovered the conventional syntax is `'eventTypes[]=1&eventTypes[]=2'`, though this appears to be a framework dependent thing.

test/actions/FilterActionsTest.js
```diff
@@ -0,0 +1,180 @@
...
+        })
+    })
+
+    describe('serializeFiltersForUrl', function() {
+        const filterNoEventTypes = { incidentId: 0 }
+        const filterWithEventTypes = {incidentId: 0, eventTypes: [1, 2]}
+        const filterWithEverything = {incidentId: 0, eventTypes: [1, 2], foo: 'bar'}
+
+        const expectedFilterTokensWithoutEventTypes = '?'
+        const expectedFilterTokensWithEventTypes = '?eventTypes=1&eventTypes=2'
```
strand: Use describe blocks to isolate the variable definitions to the scope of the tests in which their values are consumed.

src/reducers/eventTypeReducer.js
```diff
@@ -14,6 +14,8 @@ export const records = (state = defaultEventTypeCollection, action) => {
     switch(action.type){
         case eventTypeActions.GET_EVENT_TYPE_SUCCESS:
             return mergeToStateById(state, action.eventType)
+        case eventTypeActions.GET_EVENT_TYPES_SUCCESS:
```
strand: Use the multiple case for a single op syntax: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#Multi-case_-_single_operation

test/actions/EventActionsTest.js
```diff
@@ -0,0 +1,7 @@
+'use strict'
```
strand: Delete empty test files.

test/actions/FilterActionsTest.js
```diff
@@ -0,0 +1,180 @@
+'use strict'
+import { expect } from 'chai'
+import queryString from 'query-string'
+import * as filterActions from '../../src/actions/filterActions'
+
+describe('FilterActions', function () {
+    describe('getFilterFromUrl', function() {
```
strand: I want to pair with you on testing, there are some good opportunities to improve clarity  in your testing strategy.

test/actions/FilterActionsTest.js
```diff
@@ -0,0 +1,180 @@
...
+            const testQueryTwo = '?eventTypes=1'
+            const resultTwo = filterActions.getFilterFromUrl(testQueryTwo)
+
+            expect(resultTwo).to.be.an('object')
+            expect(resultTwo.eventTypes).to.be.an('array')
+        })
+        it('should return an int array', function() {
+            const unexpectedValue = ["16", "11"]
+
+            expect(resultOne.eventTypes[0]).to.not.equal(unexpectedValue[0])
```
strand: pseudocode expectations:

typeof(resultOne.eventTypes[0]) is a Integer
expect(resultOne.eventTypes[0]).to.equal(16)

src/actions/filterActions.js
```diff
@@ -0,0 +1,120 @@
...
+    return {
+        type: CHANGE_EVENT_FILTER,
+        filter
+    }
+}
+
+export const addFilter = (history) => (filter, eventType) => {
+    let newFilter = {}
+    let oldFilter = filter
+    if (!eventType && !eventType.id) {
```
macrobug: This will crash if eventType is null or undefined. Maybe you meant || instead of &&?

src/components/Timeline/EventFilter.js
```diff
@@ -1,22 +1,107 @@
...
+      key={eventType.id}
+      onRequestDelete={() => dispatch(filterActions.removeFilter(history)(filter, eventType))}
+      style={chipStyles.chip}
+    >
+      {eventType.name}
+    </Chip>
+  )
+}
+
+const extractEventTypesFromEventTypesObject = (eventTypes) => {
```
macrobug: Avoid using "for ... in" (see https://stackoverflow.com/questions/500504/why-is-using-for-in-with-array-iteration-a-bad-idea
) and avoid using "var" (see https://stackoverflow.com/questions/762011/whats-the-difference-between-using-let-and-var-to-declare-a-variable).

This method can be rewritten using .map() instead of a loop.

RedSquirrelious: It's not an array.  It's going through and pulling items from an object.  I don't think "map" works with that.  Do you know of a better way to iterate through an object?

src/reducers/filterReducer.js
```diff
@@ -0,0 +1,12 @@
+import * as filterActions from '../actions/filterActions'
+
+export const filter = (defaultFilter) => (state = defaultFilter, action) => {
+    switch (action.type) {
+        case filterActions.CHANGE_EVENT_FILTER:
+            return Object.assign({ eventTypes: [], ticketId: null }, action.filter)
```
macrobug: Use spread operator (...) to maintain consistency with other reducers.

test/reducers/eventReducerTest.js
```diff
@@ -0,0 +1,50 @@
+'use strict'
+import { expect } from 'chai'
+// import * as eventActions from '../../src/actions/eventActions.js'
```
macrobug: Avoid pushing dead (commented out) code.

src/reducers/eventReducer.js
```diff
@@ -11,6 +13,15 @@ const makeSearchable = (event) => ({
     filterableIncidentId: event.incidentId.toString()
 })
 
+const pageArgs = {
+    defaultPage: 1,
+    defaultSortOrder: 'desc',
+    defaultSortBy: 'occurred',
+    defaultPer: 10,
```
macrobug: what does "Per" means in this context?

src/components/Timeline/EventFilter.js
strand: You can map an object using Underscore: http://underscorejs.org/#map and I bet there's some utilities built into React for this as well.

src/actions/eventTypeActions.js
```diff
@@ -4,13 +4,18 @@ export const TRY_GET_EVENT_TYPE = 'TRY_GET_EVENT_TYPE'
...
+export const GET_EVENT_TYPES_SUCCESS = 'GET_EVENT_TYPES_SUCCESS'
+export const GET_EVENT_TYPES_FAILURE = 'GET_EVENT_TYPES_FAILURE'
 
 export const fetchEventType = (eventTypeId) => reduxBackedPromise(
     ['eventTypes/' + eventTypeId],
     getEventTypeActionSet(eventTypeId)
 )
 
-
+export const fetchEventTypes = () => reduxBackedPromise(
```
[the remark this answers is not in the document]
pdimitratos: There's a distinction between fetchEventType and fetchEventTypes, though fetch and fetchAll would work.

That said, I somewhat prefer:
import { fetchEventTypes } from '$path/eventTypeActions'

dispatch(fetchEventTypes())

to:
import * as eventTypeActions from '$path/eventTypeActions'

dispatch(eventTypeActions.fetchAll())

src/components/Timeline/EventFilter.js
pdimitratos: extractEventTypesFromEventTypesObject(eventTypes) can be replaced by Object.values(eventTypes)

src/reducers/eventReducer.js
pdimitratos: Number of records per page. The naming convention comes from the pagination middleware we're using; I don't think we can change it.

test/actions/FilterActionsTest.js
maggiepint: That started in PHP and is not guaranteed to work in other places. The Moment.js team once had a really long discussion about it for gosh-knows-what-reason.

src/reducers/eventTypeReducer.js
RedSquirrelious: These do different things.    One works with a single eventType and the other works with multiple eventTypes.

test/reducers/incidentReducersTest.js
```diff
@@ -9,7 +9,7 @@ const dummyEventActions = {
         type: 'DUMMY_FETCH_EVENTS'
     }),
 
-    getEventsActionSet: (incidentId) => ({
```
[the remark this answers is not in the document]
RedSquirrelious: Not sure and reverted.

src/actions/filterActions.js
RedSquirrelious: Won't the return be triggered with the "!eventType" if the eventType is null?

src/components/Timeline/EventFilter.js
RedSquirrelious: done

src/reducers/filterReducer.js
RedSquirrelious: done

PR #52, opened 2017-12-29
card expandable only when extra information is available

yingzhao-1:


src/components/Timeline/Event.js
```diff
@@ -24,37 +24,38 @@ export const Event = ({
         animationName: 'yellowfade',
         animationDuration: '30s',
         animationDelay: -(moment().diff(event.timeReceived, 'seconds')) + 's'
-    } : {}
+  } : {}
+  const expand = (eventId || eventTypeId || ticketId || incidentId) ? true : false
```
pdimitratos: We generally want all of these (eventId, eventTypeId, ticketId, and incidentId) to be present before trying to show actions.

Also, the ternary operator is redundant, as the result of (eventId || eventTypeId || ticketId || incidentId) is already Boolean.

const expand = (eventId && eventTypeId && ticketId && incidentId)
^ may be a better version.

Also, with regard to naming, since this bool is passed to things such as actAsExpander and showExpandableButton, the name 'expand' doesn't add much information. A more informative name might relate to what we're doing to decide whether things should be expanded. Changing 'expand' to 'isAllPlaybookInfoAvailable' could add clarity

PR #53, opened 2017-12-29
Sticky Extension State

maggiepint:
Pull request persists some of the redux state to local storage, allowing us to return the user to the ticket they were looking at before, unless the request a different ticket.

This change stages some awesome possibilities including:
List of recently visited tickets
Sticky filters

Over time, we will need to prune some of the retained state to avoid blowing the browser out of memory - but this will do for an MVP

### DIFF
--- a/src/components/Home.js
+++ b/src/components/Home.js
@@ -1,0 +1,25 @@
+import React from 'react'
+import { connect } from 'react-redux'
+import { Redirect } from 'react-router'
+
+
+
+export const Home = ({ ticket }) => {
+    if (ticket && ticket.originId) {
+        return <Redirect to={`/tickets/${ticket.originId}`} />
+    }
+    return <Redirect to={'/search'} />
+}
+
+export const mapStateToProps = (state) => {
+    // this gross thing finds all actual tickets, omitting the refresh metadata, and then picks the most recently visited one out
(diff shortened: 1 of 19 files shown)


src/components/Home.js
```diff
@@ -0,0 +1,25 @@
...
+
+export const Home = ({ ticket }) => {
+    if (ticket && ticket.originId) {
+        return <Redirect to={`/tickets/${ticket.originId}`} />
+    }
+    return <Redirect to={'/search'} />
+}
+
+export const mapStateToProps = (state) => {
+    // this gross thing finds all actual tickets, omitting the refresh metadata, and then picks the most recently visited one out
```
pdimitratos: Eventually, I'd prefer to keep track of this by dispatching something to a reducer when the user loads a new incident, since this implementation keeps us from doing things like refreshing incidents in the background every few minutes.

src/components/Home.js
```diff
@@ -0,0 +1,25 @@
...
+    // the sort function works because the dates are UTC and thus lexically sortable to start with
+    return {
+        ticket: (state.tickets && state.tickets.map) ? Object.values(state.tickets.map).filter(ele => { return ele.lastRefresh })
```
pdimitratos: Does the filter work as just .filter(ele => ele.lastRefresh)? lastRefresh should be truthy, right?

PR #56, opened 2018-01-03
code hygiene:  propTypes in components

RedSquirrelious:


test/services/PlaybookServiceTest.js
```diff
@@ -37,7 +37,10 @@ describe('Playbook Service', function () {
                 dateTimeComparisonValue: '19700101',
                 dataFormat: dateTimeDataFormatValue
             })
-            expect(playbook.GetComparisonValue(getComparisonValueDateTime).isSame('19700101')).to.be.true
+
```
strand: I don't see why this revision needs to occur within this propType work. Is this related? If not, please restore to the master version.

test/components/Auth/EnsureLoggedInTest.js
```diff
@@ -47,4 +49,78 @@ describe('EnsureLoggedIn', function test () {
     it('Should render children when no error is present and user is logged in', () => {
         expect(this.loggedInOutput.type).to.equal('div')
     })
+
```
strand: There's a missing context here, which I can tell because there is shared setup between two context or describe blocks. I believe its title is 'when the console has an error'

src/reducers/index.js
```diff
@@ -1,3 +1,4 @@
+import { combineReducers } from 'redux'
```
strand: I don't see why this revision needs to occur within this propType work. Is this related? If not, please restore to the master version.

src/configureStore.js
```diff
@@ -1,4 +1,4 @@
-import { createStore, applyMiddleware, compose, combineReducers } from 'redux'
+import { createStore, applyMiddleware, compose } from 'redux'
```
strand: I don't see why this revision needs to occur within this propType work. Is this related? If not, please restore to the master version.

src/components/Engagements.js
```diff
@@ -17,7 +17,7 @@ export const mapStateToPropsEngagements = (state, ownProps) => {
     }
 }
 
-export const Engagements = ({dispatch, incidentId, engagements, user}) => {
+export const Engagements = ({ dispatch, incidentId, engagements, user }) => {
```
strand: I don't see why this revision needs to occur within this propType work. Is this related? If not, please restore to the master version.

src/actions/incidentActions.js
```diff
@@ -151,5 +151,5 @@ export const updateIncidentCreationInput = (input) => ({
 
 export const duplicateIncident = (ticketId) => (dispatch) => {
     dispatch(createIncidentActionSet(ticketId, {}).fail({ message: 'An incident already exists for this ticket'}))
-    dispatch(ticketActions.updateIncidentQuery(ticketId))
+    dispatch(ticketActions.updateTicketQuery(ticketId))
```
strand: I don't see why this revision needs to occur within this propType work. Is this related? If not, please restore to the master version.

test/components/Auth/EnsureLoggedInTest.js
```diff
@@ -47,4 +49,78 @@ describe('EnsureLoggedIn', function test () {
...
+        let correctType = {
+            error: 'TESTTEST',
+            isLoggedIn: true,
+            store: GetMockStore({ auth: { error: 'Test Error' } })
+        }
+
+        describe('correct propTypes', function () {
+            it('should return false for no error with the propType', function () {
+                let result = setup(correctType, children)
+                let p = isPropTypeWarningMatch(console.error.args.toString())
```
strand: Never use single letter variables. `hasWarning` would be a fine name for this.

test/test-utils.js
```diff
@@ -0,0 +1,17 @@
+import sinon from 'sinon'
+
+export const isPropTypeWarningMatch = (warning) => {
+    return typeof(warning) === 'string' &&
+        warning.indexOf('Warning: Failed prop type:') > -1
+}
+
+
+export const stubConsoleError = () => {
```
strand: Don't abstract before/after blocks from their test context, this is important info which should be within your specs.

test/components/Auth/EnsureLoggedInTest.js
```diff
@@ -47,4 +49,78 @@ describe('EnsureLoggedIn', function test () {
...
+
+                sinon.assert.notCalled(console.error)
+                expect(p).to.be.false
+            })
+        })
+    })
+
+    context('propTypes passed in wrong', function() {
```
strand: As far as I can tell, these specs test the propTypes modules implementation. When testing, only test code you control. You aren't a propType maintainer, so don't test the behavior proTyprs provide.

src/actions/incidentActions.js
RedSquirrelious: There is no updateIncidentQuery in ticketActions.  Reverted back to master for now.  Can be cleaned up later.

src/components/Engagements.js
RedSquirrelious: No problem.  Reverted

src/configureStore.js
RedSquirrelious: combineReducers isn't used, resulting in linter errors.  Reverted to master to help clarify this PR.

src/reducers/index.js
RedSquirrelious: It was missing and causing linting errors.  It isn't needed because the filter reducer is combined with the event reducer in the event reducer file.  Someone else already fixed it on master.  Reverted.

test/components/Auth/EnsureLoggedInTest.js
RedSquirrelious: The consensus seems to be that we are not going to test for propTypes.  Tests removed accordingly.  :)

src/configureStore.js
RedSquirrelious: Actually, someone removed combineReducers already from master.  Not sure why this is showing us as a difference.

RedSquirrelious:
Branch had some odd issues with master.  Strand rebased a version on their computer, which fixed the issue.  See PR # 57.

PR #57, opened 2018-01-03
Features/prop types all around cleanup

strand:


strand:
#56 with extraneous files removed. TIL that if a branch hasn't been rebased off of master in a while, that `git checkout -- FILENAME` can result in a no-op instead of restoring the content of that file from master.

strand:
@RedSquirrelious This gets rid of the weird conflicts we were seeing… feel free to merge at will.

PR #59, opened 2018-01-04
filter chips pulled into separate component

RedSquirrelious:


src/components/elements/FilterChips.js
```diff
@@ -0,0 +1,75 @@
...
+    filter: PropTypes.object,
+    filterType: PropTypes.string,
+    filterOptions: PropTypes.object,
+    lookupMethod: PropTypes.func,
+}
+
+export const mapStateToProps = (state, ownProps) => {
+    return {
+        ...ownProps,
+        filter: state.events.filter ? state.events.filter : {},
```
pdimitratos: Reducers should always exist; if they don't, there's a major problem and we should want an error to occur so that we're not silently failing to find a path that's been moved or mistyped, in my opinion.

src/components/elements/FilterChips.js
```diff
@@ -0,0 +1,75 @@
...
+            </div>
+        )
+    }
+    else {
+        return <div></div>
+    }
+
+}
+
+const renderChip = (dispatch, history, filter, filterOptions, selectedFilterType, lookupMethod) => {
```
pdimitratos: This method would be more focused if it only took three args: The id, the name, and a function that when called with no arguments dispatches the (already populated) removeFilter action. Separate the concern of actually rendering the chip from the concern of transforming the data.

src/components/elements/FilterChips.js
```diff
@@ -0,0 +1,58 @@
+import React from 'react'
+import { connect } from 'react-redux'
+import PropTypes from 'prop-types'
+import Chip from 'material-ui/Chip'
+import ByPath from 'object-path'
+
+export const FilterChips = ({filter, selectSpecificFilter, records, onRequestDelete}) => {
+    let selectedFilters = ByPath.get(filter, selectSpecificFilter)
```
pdimitratos: This should be const since we don't change it

pdimitratos: This should be const since we don't change it anywhere.

src/components/elements/FilterChips.js
```diff
@@ -0,0 +1,58 @@
...
+    return selectedFilters? selectedFilters.map((selectedFilter) => hydrateChip(selectedFilter, records))
```
pdimitratos:
```
return selectedFilters
    ? selectedFilters
        .map()
        .map()
    : <div></div>
```

is my preferred format for this. Lining up the map statements makes it more clear that it's a pipeline rather than a map called within another map call. Lining up the ternary operator increases uniformity with the rest of the codebase.

src/components/elements/FilterChips.js
```diff
@@ -0,0 +1,58 @@
...
+        .map(chip => renderChip(chip.id, chip.name, onRequestDelete(filter, chip.id)))
+        :
+        <div></div>
+}
+
+export const mapStateToProps = (state, ownProps) => {
```
pdimitratos: This there's no body here, this can use the same format as hydrateChip for just returning an object.

src/components/elements/FilterChips.js
```diff
@@ -0,0 +1,58 @@
...
+FilterChips.propTypes = {
+    selectSpecificFilter: PropTypes.string,
+    lookupFilterObject: PropTypes.string,
+    recordLookup: PropTypes.string,
+    onRequestDelete: PropTypes.func
+}
+
+
+
+export default connect(mapStateToProps)(FilterChips)
```
pdimitratos: Ending files with newline characters is preferred for git diff compatibility; git uses a red complaint icon at end of file on pull requests when the file doesn't have one. We should probably add this to our linter rules.

src/components/Timeline/EventFilter.js
```diff
@@ -31,11 +31,15 @@ export const filterSearchForm = {
...
-      {filterChips}
+      <FilterChips
+        selectSpecificFilter={'eventTypes'}
+        lookupFilterObject={'events.filter'}
+        recordLookup={'eventTypes.records'}
+        onRequestDelete={(filter, id) => () => dispatch(filterActions.removeFilter(history, 'eventTypes')(filter,id))}
+      />
       <AutoComplete
         floatingLabelText="Filter by event type"
         filter={AutoComplete.caseInsensitiveFilter}
```
pdimitratos: There's still a bug in onNewRequest, because the object passed into the function can be one of two things: when someone clicks on the list, it gets the eventType object selected (which works), but when someone presses enter, it gets just the text (example: 'Note'), and dispatch fails.
  
I suggest pulling out the onNewRequest function into a defined const with a signature like 

const onNewRequest = (filterTypes) => (inputResult) => {
Which makes the argument to AutoComplete

onNewRequest={onNewRequest(filterTypes)}

Then in onNewRequest you can check whether inputResult is an object or a string, and if it's a string you can use the string to lookup the eventType by name in filterTypes.

pdimitratos: Actually, probably better to fix this one in https://github.com/Azure/Sia-EventUI/pull/60 , so no changes needed for that on this pull request

RedSquirrelious:
@pdimitratos changes made per your request.

PR #60, opened 2018-01-04
pulled filter autocomplete menu into separate component

RedSquirrelious:


src/components/elements/AutoCompleteMenu.js
```diff
@@ -0,0 +1,35 @@
...
+    let filterSearchForm = menuFilterSearchForm
+    return filter? (
+        <div>
+            <AutoComplete
+                floatingLabelText={label}
+                filter={AutoComplete.caseInsensitiveFilter}
+                dataSource={menuOptions}
+                searchText={filterSearchField}
+                onUpdateInput={(searchText) => dispatch(formActions.updateInput(filterSearchForm.name, filterSearchForm.field, searchText))}
+                onNewRequest={
```
pdimitratos: There's still a bug in onNewRequest, because the object passed into the function can be one of two things: when someone clicks on the list, it gets the eventType object selected (which works), but when someone presses enter, it gets just the text (example: 'Note'), and dispatch fails.

I suggest pulling out the onNewRequest function into a defined const with a signature like

const onNewRequest = (filterTypes) => (inputResult) => {

Which makes the argument to AutoComplete:

onNewRequest={onNewRequest(filterTypes)}

Then in onNewRequest you can check whether inputResult is an object or a string, and if it's a string you can use the string to lookup the eventType by name in filterTypes.

src/components/Timeline/EventFilter.js
```diff
@@ -30,25 +23,20 @@ export const filterSearchForm = {
   field: 'input'
 }
 
-
 const EventFilter = ({pagination, filter, filterSearchField, eventTypes, filterTypes, dispatch, history}) =>  {
   const filterChips = filter && filter.eventTypes && eventTypes ? renderChips(history, filter, eventTypes, dispatch): null
+  // debugger
```
strand: Remove comments

src/components/elements/AutoCompleteMenu.js
```diff
@@ -0,0 +1,35 @@
...
+                }
+                dataSourceConfig={dataSourceConfig}
+            />
+        </div>
+    )
+    : (
+        <div></div>
+    )
+}
+export default AutoCompleteMenu
```
strand: Where are the specs?

RedSquirrelious:
Taken care of in PR #70

PR #61, opened 2018-01-04
Disable iconButton when there is no extra information about the event

yingzhao-1:


src/components/Timeline/Event.js
```diff
@@ -20,13 +21,16 @@ export const Event = ({
...
-    event
+    event,
+    actions
 }) => {
     const itemHighlight = (event && event.timeReceived) ? {
         animationName: 'yellowfade',
         animationDuration: '30s',
         animationDelay: -(moment().diff(event.timeReceived, 'seconds')) + 's'
     } : {}
+    const isAllPlaybookInfoAvailable = actions.length > 0
```
strand: I think this should be renamed to `isPlaybookAvailable` (I don't believe this test case covers _all_, but instead _any_ playbook availability, and info is meaningless.)

src/components/Timeline/Event.js
```diff
@@ -42,22 +46,30 @@ export const Event = ({
         <Card
             className="incident-card"
             style={{ backgroundColor }}
+            expanded={isAllPlaybookInfoAvailable ? null : false}
```
strand: How does the expanded property being null differ from false within the view context? Does this change what is displayed? I would much prefer to use something that evals to true or false over something which evals to null or false as nil/null. See also: http://www.yegor256.com/2014/05/13/why-null-is-bad.html

(Specifically, this is confusing as heck for folks who rely on falsiness or truthiness in conditionals, as `!!false == !!null`.) I think collapsible is the antonym of expandable.

src/components/Timeline/Event.js
```diff
@@ -20,13 +21,16 @@ export const Event = ({
...
+    event,
+    actions
 }) => {
     const itemHighlight = (event && event.timeReceived) ? {
         animationName: 'yellowfade',
         animationDuration: '30s',
         animationDelay: -(moment().diff(event.timeReceived, 'seconds')) + 's'
     } : {}
+    const isAllPlaybookInfoAvailable = actions.length > 0
+    const iconColor = isAllPlaybookInfoAvailable ? 'black' : 'Lightgrey'
```
strand: Inline this to the context of the color attribute.

src/components/Timeline/Event.js
```diff
@@ -76,10 +88,20 @@ Event.propTypes = {
...
 
 export const mapStateToEventProps = (state, ownProps) => {
-    const event = ownProps.event
-    const eventType = state.eventTypes.records[event.eventTypeId]
-    const ticket = state.tickets.map[ownProps.ticketId]
-    const engagement = state.engagements.list.find(engagement => engagement.id === ownProps.engagementId)
+  const event = ownProps.event
+  const eventType = state.eventTypes.records[event.eventTypeId]
+  const ticket = state.tickets.map[ownProps.ticketId]
+  const engagement = state.engagements.list.find(engagement => engagement.id === ownProps.engagementId)
```
pdimitratos: Event is not called with an engagementId. I know this isn't something you introduced, but it does cause a problem for determining which actions are qualified (for the ones that depend on the engagement). Hoisting the engagement selection to mapStateToDisplayPlaybookProps and passing engagementId alongside eventId, eventTypeId, ticketId, and incidentId though Playbook to DisplayPlaybook simplifies DisplayPlaybook significantly.

src/components/Timeline/Playbook/DisplayPlaybook.js
```diff
@@ -59,20 +52,9 @@ export const mapStateToDisplayPlaybookProps = (state, ownProps) => {
...
-                ? populatedConditionSetTest(currentConditionSet)
-                : false,
-            true
-        )
-    )
+  
 
     return {
-        actions: qualifiedActions,
         engagementId: engagement ? engagement.id : null,
```
pdimitratos: Since the only thing that gets pulled out of state here now is engagementId, and we should pull that out when we're getting qualifiedActions, this mapStateToProps function doesn't need to exist; we can pass everything we need for DisplayPlaybook from Event.js through Playbook.js.

src/components/Timeline/Event.js
pdimitratos: The MaterialUI Card controls have this weird behavior where you can control expansion manually by passing true or false or let the Card control expansion in local state by passing null. We want to be able to prevent expansion at this level but not explicitly control when things are expanded, so this is valid within the context of how Card handles these arguments. Whether Card is well designed in the first place is a different discussion.

src/components/Timeline/Event.js
```diff
@@ -42,22 +46,30 @@ export const Event = ({
         <Card
             className="incident-card"
             style={{ backgroundColor }}
+            expanded={isAllPlaybookInfoAvailable ? null : false}
         >
             <CardHeader
                 title={ticketId ? `${ticketId}: ${text}` : text}
                 subtitle={time ? time.local().format('LTS') : 'Time unknown!'}
                 actAsExpander={true}
                 showExpandableButton={true}
```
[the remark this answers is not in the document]
yingzhao-1: so you prefer your version?

src/components/Timeline/Event.js
strand: tire fire on a tire swing

PR #62, opened 2018-01-05
fix persisting error eventTypeReducer

RedSquirrelious:
use of persistCombineReducers caused the eventType errors to hang around and cause display problems.

### DIFF
--- a/src/reducers/eventTypeReducer.js
+++ b/src/reducers/eventTypeReducer.js
@@ -1,8 +1,9 @@
+import { combineReducers } from 'redux'
 import * as eventTypeActions from '../actions/eventTypeActions'
 import { mergeToStateById } from './reducerHelpers/merge'
 import buildFetching from './reducerHelpers/fetching'
 import buildError from './reducerHelpers/error'
-import { persistCombineReducers } from 'redux-persist'
+import { persistReducer } from 'redux-persist'
 import storage from 'redux-persist/lib/storage' // default: localStorage if web, AsyncStorage if react-native
 
 const persistConfig = {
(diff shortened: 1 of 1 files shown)


src/reducers/eventTypeReducer.js
```diff
@@ -33,10 +34,12 @@ export const records = (state = defaultEventTypeCollection, action) => {
...
 }
 
-export const eventTypeReducer = persistCombineReducers(persistConfig, {
+const persistedRecords = persistReducer(persistConfig, records)
+
+export const eventTypeReducer = combineReducers({
     fetching,
     error,
-    records
+    records: persistedRecords
```
strand: What bug was occurring before this persistence patch was made? Is it something which we could add a test around (with any sort of regression or missed expectation I prefer to write a test for the expected behavior, so as to demonstrate what is being fixed and prevent it from breaking again.)

pdimitratos: Fetching errors in one session would persist in the next session, so once an eventType failed to load once, the UI would never even try to load it without manual retry.

RedSquirrelious: Right now we don't have a way to test this.  We only wanted one reducer to persist but used the method that caused all reducers to persist.  This sets it to only keep the reducer we want.  (the problem we were having was that rather than display the events, it showed an error message on the page because it thought we still had an error)

strand: 🤔 I think it is ok to not spin our wheels trying to test this, as we are relying on a third-party library for the intended behavior. Because we are depending on the `persistReducer` and it depends on the `persistConfig` I'd like these to be clumped together in the same area of the code… You can either move the config to the same area as the persistReducer instantiation, or inline the logic to the scope of the eventTypeReducer. I slightly prefer inlining, which would look like `persistReducer({ key: 'eventType', storage }, records)`. I prefer this because I only need to read one line of code to know that other people's code is responsible for persisting our records.

PR #64, opened 2018-01-16
43308: Remove alias from sandwich menu dropdown

yingzhao-1:


src/components/TopNav/NavMenu.js
```diff
@@ -8,26 +8,20 @@ import NavigationMenu from 'material-ui/svg-icons/navigation/menu'
...
 
-export const mapStateToProps = (state) => {
-    return {
-        alias: state.auth.userAlias
-    }
-}
 
-export const NavMenu = ({dispatch, alias}) =>
+
+export const NavMenu = dispatch =>
```
pdimitratos: Changing from a [destructuring function signature](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) to a plain args signature is almost always a breaking change; this version of NavMenu will receive as an argument an object named 'dispatch' with a property 'dispatch.dispatch' that's a function. That breaks the LogOut link.

`export const NavMenu = ({dispatch}) =>`

will fix that issue.

src/components/TopNav/NavMenu.js
```diff
@@ -8,26 +8,24 @@ import NavigationMenu from 'material-ui/svg-icons/navigation/menu'
...
     targetOrigin={{horizontal: 'left', vertical: 'top'}}
 >
-    <MenuItem primaryText={alias} />
-    <MenuItem primaryText={<Link to='/search' >Incident Search</Link>} />
-    <MenuItem primaryText={<Link to='/' onClick={() => dispatch(auth.logOut)}>LogOut</Link>} />
-    <MenuItem primaryText={<Link to='/debug' >Debug</Link>} />
-  </IconMenu>
+
+    <MenuItem primaryText={<Link to="/search" >Incident Search</Link>} />
+    <MenuItem primaryText={<Link to="/" onClick={() => dispatch(auth.logOut)}>LogOut</Link>} />
```
RedSquirrelious: This all looks good.  I approve the commit.  Please feel free to merge.  

On a side note, I am curious about the switch from single quotes to double quotes.  Are we doing it that way now?  Was it necessary for some reason?  Or is it personal preference?  :)

macrobug: This breaks the linter. Should be single quotes.

PR #65, opened 2018-01-17
Code Maintenance

macrobug:
- Standard JS + all linter fixes
- Code Coverage metrics
- Clean imports and dependencies
- Refactored config
- Readme badges


cfg/defaultConstants.js
```diff
@@ -14,7 +16,7 @@ module.exports = {
 
   ticketRefreshIntervalInSeconds: 300,
 
-  authVersion: 'ADAL',
+  authVersion: 'TEST',
```
pdimitratos: Not sure if we want to do it this way; the default for running localhost should be ADAL. We only want TEST when actually running tests; the app should be completely non-functional when authVersion is TEST. Having to override configuration before launching the app for the first time gets in the way of 'clone and it just works' as a low barrier to entry on the project.

src/actions/actionHelpers.js
```diff
@@ -35,39 +37,39 @@ const sortActionType = (BASE_NAME) => 'SORT_' + BASE_NAME
...
-        filter
-    })
+  filter: (filter) => ({
+    type: filterActionType(BASE_NAME),
+    filter
+  })
 })
 
-/*Update pagination does nothing on its own, but activates
+/* Update pagination does nothing on its own, but activates
```
pdimitratos: Is the spacing for comments part of standardJS, or a matter of preference? I don't mind it, just curious.

src/actions/engagementActions.js
```diff
@@ -1,5 +1,5 @@
 import moment from 'moment'
-import { reduxBackedPromise, paginationActions } from './actionHelpers'
+import { reduxBackedPromise, paginationActions } from 'actions/actionHelpers'
```
pdimitratos: I really like this whole alias thing. Makes me wonder what other aliases would be useful; how much is too much?

src/actions/engagementActions.js
```diff
@@ -9,72 +9,70 @@ export const DISENGAGE_SUCCESS = 'DISENGAGE_SUCCESS'
...
+
 export const disengage = (participant, engagement, timeDisengaged = moment()) =>
 reduxBackedPromise(
-    [
-        'incidents/' + engagement.incidentId + '/engagements/' + engagement.id,
-        updatedEngagement(engagement, timeDisengaged),
-        null,
-        false
-    ],
+  [
```
pdimitratos: Does this violate the 'never start a line with '[' constraint?

src/actions/eventActions.js
```diff
@@ -29,122 +28,120 @@ export const fetchEvents = (filter) => reduxBackedPromise(
...
+  try: () => ({
+    type: REQUEST_EVENT,
+    incidentId,
+    id: eventId
+  }),
+
+  succeed: (event) => (dispatch) => {
+    dispatch({
+      type: RECEIVE_EVENT,
+      event,
```
pdimitratos: I just realized that 'event' is a reserved word in JavaScript. Do you think this could cause issues during local debugging? (I assume it would be minified away in the dist version).

src/components/Home.js
```diff
@@ -2,24 +2,21 @@ import React from 'react'
...
+  return <Redirect to={'/search'} />
 }
 
 export const mapStateToProps = (state) => {
     // this gross thing finds all actual tickets, omitting the refresh metadata, and then picks the most recently visited one out
     // the sort function works because the dates are UTC and thus lexically sortable to start with
-    return {
-        ticket: (state.tickets && state.tickets.map) ? Object.values(state.tickets.map).filter(ele => { return ele.lastRefresh })
+  return {
+    ticket: (state.tickets && state.tickets.map) ? Object.values(state.tickets.map).filter(ele => { return ele.lastRefresh })
```
pdimitratos: This makes me wonder what sort of rules we could add related to use of the ternary operator. I like

`condition ? doIfTrue() : doIfFalse()`

iff it all fits on one line, but if it has to be split into multiple lines, I usually find

`condition`

`  ? doIfTrue()`

`  : doIfFalse()`

if it's too long for one line. What do you think?

cfg/defaultConstants.js
macrobug: synced offline.

src/components/Home.js
macrobug: Agreed. Will look into a lint rule for this.

src/actions/eventActions.js
macrobug: Seems it's not a reserved word, but a global variable in some browsers. May be a good idea to change it.

PR #70, opened 2018-01-23
Feature/filters autocomplete

RedSquirrelious:
@strand @macrobug  Please take a look.....   this should make the autocomplete menu more generic.  there are tests.


test/components/elements/AutoCompleteMenuTest.js
```diff
@@ -0,0 +1,227 @@
...
+
+      it('should return a div', function () {
+          expect(this.output.type).to.equal('div')
+      })
+
+      it('the div should contain an AutoComplete', function () {
+          expect(this.output.props.children.type).to.equal(AutoComplete)
+      })
+
+      it('the Autocomplete should have the label we passed in', function () {
```
macrobug: maybe reword to `it('should.........?` and put the `'Autocomplete'` in a `describe()`

test/components/elements/AutoCompleteMenuTest.js
```diff
@@ -0,0 +1,227 @@
...
+
+      it('the Autocomplete should have an onUpdateInput fx just like the fx we passed in', function () {
+          expect(this.output.props.children.props.onUpdateInput()).to.equal(mockProps.onUpdateInput())
+      })
+    })
+  })
+
+  describe('onNewRequest', function() {
+    const selectMethod = (input) => input
+    const clearMethod = () => console.log('')
```
macrobug: is this clearMethod like a no-op? the test runner may not necessarily have a console object, and even if it does, this may have side effects.

test/components/elements/AutoCompleteMenuTest.js
```diff
@@ -0,0 +1,227 @@
...
+        it('it should call the clear method once', function () {
+          expect(clearSpy).to.have.been.called.once
+        })
+
+        it('should not call the select method', function () {
+          expect(selectSpy).to.not.have.been.called
+        })
+      })
+
+      describe('when the data source has invalid data', function () {
```
strand: How is this data source invalid?

test/components/elements/AutoCompleteMenuTest.js
```diff
@@ -0,0 +1,227 @@
...
+        })
+
+        it('should call the select method once', function () {
+          expect(selectSpy).to.have.been.called.once
+        })
+      })
+
+    })
+
+    context('when the input is invalid', function() {
```
strand: I think this context is not _invalid_, but instead _when a clear autocomplete match can not be found_ and this context's sibling would be _when a clear autocomplete match can be found_.

test/components/elements/AutoCompleteMenuTest.js
RedSquirrelious: The method is looking for "id" and the data has "title" instead.  I'll clarify in the describe.

PR #73, opened 2018-01-26
display eventType fetching errors w/formatting 

RedSquirrelious:


src/components/Timeline/Event.js
```diff
@@ -30,11 +30,11 @@ export const Event = ({
     animationDelay: -(moment().diff(event.timeReceived, 'seconds')) + 's'
   } : {}
   const isAllPlaybookInfoAvailable = !!(actions && Array.isArray(actions) && actions.length > 0)
-
+  const missingId = eventTypeId
   return eventTypeIsFetching && !eventHasValidDisplayText(event)
         ? LoadingMessage('Fetching Event Type Information', eventTypeActions.fetchEventType(eventTypeId))
         : eventTypeIsError && !eventHasValidDisplayText(event)
-            ? ErrorMessage('Error fetching eventType!', eventTypeActions.fetchEventType(eventTypeId))
+      ? ErrorMessage(`Error fetching eventType: ${missingId}`, eventTypeActions.fetchEventType(eventTypeId), time, backgroundColor)
```
pdimitratos: Indentation on this doesn't make it clear that it's dependent on line 36; alternatives in a ternary should be aligned vertically to make it more clear which cases are paired.

src/components/elements/ErrorMessage.js
```diff
@@ -1,11 +1,20 @@
...
-  { actionForRetry ? <RetryButton actionForRetry={actionForRetry} /> : null }
+const ErrorMessage = (message, actionForRetry, time = null, backgroundColor = null) => <div>
+  <Card
+    className='incident-card'
+    style={{ backgroundColor }}
+  >
+    <ErrorIcon />
+    <CardHeader
+      title={message}
+      subtitle={time ? time.local().format('LTS') : 'Time unknown!'}
```
pdimitratos: Not sure whether it's preferable to include 'Time unknown!' or to not display a subtitle when time is unknown. I don't have any strong feelings one way or the other, but it may end up being clutter we don't need on the UI.

Another option is to treat null differently from undefined; when the calling code passes in a time value, if it has value then we display the time, if it's undefined we display 'Time unknown!' (because presumably the calling code was trying to pass in a time value), and if it's null (the default case), then the calling code isn't trying to use the time subtitle, and we don't display it. That may be the best of both worlds but increase complexity. Up to you which approach to use (or to stick with this one).

test/components/elements/ErrorMessageTest.js
```diff
@@ -1,25 +1,54 @@
...
 
-    expect(testObject.props.children[2].type).to.equal(RetryButton)
-    expect(testObject.props.children[2].props.actionForRetry).to.equal('TestAction')
+    it('Should additionally render a retry button', function () {
+      expect(testObject.props.children.props.children[2].type).to.equal(RetryButton)
+      expect(testObject.props.children.props.children[2].props.actionForRetry).to.equal('TestAction')
+    })
+  })
+
+  describe('when given a message, an action, and a time', function () {
```
pdimitratos: It might be best to test in full isolation, giving just a time and expecting the subtitle, but not testing any other aspect. That's all that's covered by your expect statement, so including the message and action doesn't seem needed. You could just pass in null for the first two arguments and the rest of the test remains the same. (and the describe would state just 'when given a time', making it more clear precisely what's being tested).

Same for the next describe statement; a test of just the background color with null message and action should suffice.

src/components/elements/ErrorMessage.js
```diff
@@ -1,11 +1,20 @@
 import React from 'react'
 import ErrorIcon from 'material-ui/svg-icons/alert/error'
+import { Card, CardHeader } from 'material-ui/Card'
 import { RetryButton } from 'components/elements/Buttons'
 
-const ErrorMessage = (message, actionForRetry) => <div>
-  <ErrorIcon />
-  <span>{message}</span>
-  { actionForRetry ? <RetryButton actionForRetry={actionForRetry} /> : null }
+const ErrorMessage = (message, actionForRetry, time = null, backgroundColor = null) => <div>
```
pdimitratos: We may want to validate that time is an instance of moment before we use it, and wrap it in a moment call if it's not. @maggiepint probably knows better what the best thing to do in these situations is.

src/components/elements/ErrorMessage.js
RedSquirrelious: Let's leave it with a null subtitle for now when time is undefined, null, or not a moment instance.

PR #75, opened 2018-02-02
Provide breadcrumbs of prev visited tickets

strand:
## User Story

As a Sia user, I want to be able to navigate to previously viewed tickets.

Given I am within Sia on an incident ticket page
And I navigate to several other incident ticket pages
When I click on the hamburger dropdown menu
Then I should see a list of incidents
And that list should have all recently visited incidents
And that list should not have the current incident.

## Caveats

As implemented there is no ceiling on the number of incidents displayed within the menu and they are not ordered. Before merging you should verify that this behavior is correct, or adjust the behavior to limit the number of incidents displayed and sort them reverse chronologically by recency of visit.

### DIFF
--- a/src/components/TopNav/NavMenu.js
+++ b/src/components/TopNav/NavMenu.js
@@ -7,19 +7,44 @@ import { Link } from 'react-router-dom'
 import NavigationMenu from 'material-ui/svg-icons/navigation/menu'
 import IconButton from 'material-ui/IconButton'
 import * as auth from 'services/authNService'
+import { _ } from 'underscore'
 
-export const NavMenu = ({ dispatch }) =>
-  <IconMenu
+const Tickets = () => {
+  var persistedTickets  = JSON.parse(localStorage['persist:ticket'])
+  var currentTicketId   = window.location.pathname.match(/(\d+)/)[1]
 
+  var idIsNumericalAndNotCurrentTicket  = (id) => /\d+/.test(id) && id != currentTicketId
+  var addCurrentPageToHistory           = () => { history.push(this.to) }
+  var transformIdToTicketLink           = (id) =>
+    <MenuItem
+      key={'ticket-' + id}
+      primaryText={
+        <Link
+          to={'/tickets/' + id}
+          onClick={ addCurrentPageToHistory }>
+          Ticket {id}
+        </Link>
+      }
+    />
+
+  var ticketIds = _.select(Object.keys(persistedTickets), idIsNumericalAndNotCurrentTicket)
+
+  return ticketIds.map(transformIdToTicketLink)
+}
+
+export const NavMenu = (dispatch) => {
+  return (<IconMenu
     iconButtonElement={<IconButton><NavigationMenu /></IconButton>}
     anchorOrigin={{horizontal: 'left', vertical: 'bottom'}}
     targetOrigin={{horizontal: 'left', vertical: 'top'}}
->
+  >
+    <MenuItem key='search' primaryText={<Link to='/search' >Incident Search</Link>} />
+    <MenuItem key='logout' primaryText={<Link to='/' onClick={ () => dispatch(auth.logOut)}>LogOut</Link> } />
 
-    <MenuItem primaryText={<Link to='/search' >Incident Search</Link>} />
-    <MenuItem primaryText={<Link to='/' onClick={() => dispatch(auth.logOut)}>LogOut</Link>} />
-    <MenuItem primaryText={<Link to='/debug' >Debug</Link>} />
-  </IconMenu>
+    { Tickets() }
+    <MenuItem key='debug' primaryText={<Link to='/debug' >Debug</Link>} />
+  </IconMenu>)
+}
 
 NavMenu.propTypes = {
   dispatch: PropTypes.func


src/components/TopNav/NavMenu.js
```diff
@@ -7,19 +7,44 @@ import { Link } from 'react-router-dom'
...
+        </Link>
+      }
+    />
+
+  var ticketIds = _.select(Object.keys(persistedTickets), idIsNumericalAndNotCurrentTicket)
+
+  return ticketIds.map(transformIdToTicketLink)
+}
+
+export const NavMenu = (dispatch) => {
```
pdimitratos: ({dispatch}) => (dispatch) is a breaking change. React passes three arguments and the 'arguments' on the <NavMenu dispatch={dispatch} secondArg={secondValue} /> syntax are passed as properties on a single object passed to the component as the first argument: { dispatch:dispatch, secondArg:secondValue }. Removing the {} means NavMenu will receive a 'dispatch' object with a dispatch property whose value we want to use as the dispatch on line 42.

src/components/TopNav/NavMenu.js
```diff
@@ -7,19 +7,44 @@ import { Link } from 'react-router-dom'
...
+import { _ } from 'underscore'
 
-export const NavMenu = ({ dispatch }) =>
-  <IconMenu
+const Tickets = () => {
+  var persistedTickets  = JSON.parse(localStorage['persist:ticket'])
+  var currentTicketId   = window.location.pathname.match(/(\d+)/)[1]
 
+  var idIsNumericalAndNotCurrentTicket  = (id) => /\d+/.test(id) && id != currentTicketId
+  var addCurrentPageToHistory           = () => { history.push(this.to) }
```
pdimitratos: I think the Link component does this automatically; if it doesn't then we have other changes to make. We only add onClick for some of them to have effects other than changing the current page.

src/components/TopNav/NavMenu.js
```diff
@@ -7,19 +7,44 @@ import { Link } from 'react-router-dom'
 import NavigationMenu from 'material-ui/svg-icons/navigation/menu'
 import IconButton from 'material-ui/IconButton'
 import * as auth from 'services/authNService'
+import { _ } from 'underscore'
 
-export const NavMenu = ({ dispatch }) =>
-  <IconMenu
+const Tickets = () => {
+  var persistedTickets  = JSON.parse(localStorage['persist:ticket'])
```
pdimitratos: In general, if a component needs data, that component needs to get that data from the Redux store using a mapStateToProps function passed into connect. In this case, that would look like:

    const mapStateToNavMenuProps = (state) => {
        currentId = window.location.pathname.match(/(\d+)/)[1] 
        return {
            ticketIds: Object.keys(state.tickets.map).filter(id => id !== currentId)
        }
    }

which would change the NavMenu signature to ({ dispatch, ticketIds })

We should interact with localStorage directly only as a last resort when there's no possible way to get what we need out of the current Redux state. Debugging becomes much easier when we can rely on the fact that everything about the state of our UI is based on the state of the Redux store.

As an added opportunity, this means that the Ticket() function call in the NavMenu component becomes

    ticketIds.map(id => transformIdToTicketLink(id))

or even

    ticketIds.map(id =>
        <MenuItem key={'ticket-' + id} primaryText={<Link to={'/tickets/' + id}>Ticket {id}</Link>}/>
    )

src/components/TopNav/NavMenu.js
```diff
@@ -7,19 +7,44 @@ import { Link } from 'react-router-dom'
 import NavigationMenu from 'material-ui/svg-icons/navigation/menu'
 import IconButton from 'material-ui/IconButton'
 import * as auth from 'services/authNService'
+import { _ } from 'underscore'
 
-export const NavMenu = ({ dispatch }) =>
-  <IconMenu
+const Tickets = () => {
```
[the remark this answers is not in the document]
pdimitratos: Rather than pulling the URL from window, you can use react-router's withRouter() function to grab the history and add it to the ownProps second argument on the mapStateToProps function; we do this in EventFilter.js, though we don't pull history out of ownProps until the actual component.

As mentioned in a different comment, don't mock localStorage because we can't use it directly for React components anyway.

If you drink enough of the Redux Kool-Aid, all the functions will depend only on their arguments, which makes mocking relatively simple.

src/components/TopNav/NavMenu.js
strand: The Link component does not handle this automatically.

A Link created without this onClick will generate the correct link, but when I click through, I get an "Unexpected Failure When Attempting to Display Incident" message. This error is raise in Ticket in the code sample below:

``` javascript
    if (!incident || !incident.primaryTicket || !ticket || incident.error) {
      return UnexpectedFailureToLoadIncident()
    }
```

When a link to an incident is loaded from the navMenu, this code is triggered, with undefined for incident and an object for ticket.

src/components/TopNav/NavMenu.js
```diff
@@ -7,21 +8,39 @@ import { Link } from 'react-router-dom'
 import NavigationMenu from 'material-ui/svg-icons/navigation/menu'
 import IconButton from 'material-ui/IconButton'
 import * as auth from 'services/authNService'
+import { _ } from 'underscore'
```
pdimitratos: Don't think this import is needed anymore

test/components/TopNav/NavMenuTest.js
```diff
@@ -1,7 +1,8 @@
 'use strict'
 import { expect } from 'chai'
 import React from 'react'
-import createComponent from 'test/helpers/shallowRenderHelper'
+import { shallow } from 'enzyme'
```
strand: I investigated using Enzyme to inject state into the component using its `setState` method which only works for class-based components. I like the API which it provides, though I think these tests _could_ be implemented without it, I believe it will be useful in TDD-ing components.

test/components/TopNav/NavMenuTest.js
```diff
@@ -12,34 +13,57 @@ import { Link } from 'react-router-dom'
...
+    expect(link.props.primaryText.props.to).to.equal('/')
+    expect(link.props.primaryText.props.onClick).to.exist
+  })
 
-    it('Should render a log out link', () => {
-        expect(this.output.props.children[1].type).to.equal(MenuItem)
-        expect(this.output.props.children[1].props.primaryText.type).to.equal(Link)
-        expect(this.output.props.children[1].props.primaryText.props.to).to.equal('/')
-        expect(this.output.props.children[1].props.primaryText.props.onClick).to.exist
+  it('Should render a log out link', () => {
```
pdimitratos: May have forgotten to update the description

strand: 🤦🏻‍♀️

src/components/TopNav/NavMenu.js
pdimitratos: https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/Link.js

Looking at the source code for link, it definitely should be doing history.push(to)

PR #76, opened 2018-02-07
Yizha/application insight

yingzhao-1:


src/actions/authActions.js
```diff
@@ -3,14 +3,22 @@ export const USER_LOGGED_IN = 'USER_LOGGED_IN'
 export const USER_LOGGED_OUT = 'USER_LOGGED_OUT'
 export const USER_LOGIN_ERROR = 'USER_LOGIN_ERROR'
 
+var registerAuthenticatedId = (signInId) => {
```
pdimitratos: If this is used only within the userLoggedIn function, it might be best to define it within that function. 

Also, prefer const over var for variables when you don't intend for them to be changed after initially being defined (which is generally the case for functions), since that reduces the potential complexity of the code.

src/appInsights.js
```diff
@@ -0,0 +1,9 @@
+import config from 'config'
+
+var appInsights=window.appInsights||function(a){
+    function b(a){c[a]=function(){var b=arguments;c.queue.push(function(){c[a].apply(c,b)})}}var c={config:a},d=document,e=window;setTimeout(function(){var b=d.createElement("script");b.src=a.url||"https://az416426.vo.msecnd.net/scripts/a/ai.0.js",d.getElementsByTagName("script")[0].parentNode.appendChild(b)});try{c.cookie=d.cookie}catch(a){}c.queue=[];for(var f=["Event","Exception","Metric","PageView","Trace","Dependency"];f.length;)b("track"+f.pop());if(b("setAuthenticatedUserContext"),b("clearAuthenticatedUserContext"),b("startTrackEvent"),b("stopTrackEvent"),b("startTrackPage"),b("stopTrackPage"),b("flush"),!a.disableExceptionTracking){f="onerror",b("_"+f);var g=e[f];e[f]=function(a,b,d,e,h){var i=g&&g(a,b,d,e,h);return!0!==i&&c["_"+f](a,b,d,e,h),i}}return c
```
pdimitratos: If you wrote this, please provide comments explaining what's going on. If Application insights documentation provided this, please add a comment with a link to that documentation page. This is complex enough that I wouldn't even know where to begin trying to debug it if something causes it to break.

src/services/adalService.js
```diff
@@ -26,12 +26,15 @@ export const getAuthContext = (dispatch) => {
 export const login = (dispatch) => {
   if (typeof window !== 'undefined' && !!window) {
     getAuthContext(dispatch).login()
+
+    appInsights.trackEvent("SIA Login")
```
pdimitratos: These should be tracked in authNService rather than adalService, so we don't have to add tracking in each of adalService, msalService, and any other authentication service options we want to add. The statements here can be moved to the login and logOut methods in authNService before the switch statements to record events for all authentication types.

src/actions/authActions.js
```diff
@@ -1,16 +1,26 @@
...
 export const loginInProgress = () => ({
   type: LOGIN_IN_PROGRESS
 })
 
-export const userLoggedIn = (user) => ({
-  type: USER_LOGGED_IN,
-  user
-})
+export const userLoggedIn = (user) =>{
+  if(config.useAppInsight){
```
pdimitratos: If we're going to make the use of application insights configurable (and we should), it might be best to check the condition only once in appinsightsService, that way we don't forget to put a guard clause around one of the calls to appinsights, and that way we don't need to import config everywhere we're importing appinsights. It might even be best to wrap the appinsights service in a service similar to AuthNService wrapping ADAL to allow the use of different telemetry platforms. That may be beyond the scope of this pull request, though.

src/index.js
```diff
@@ -5,6 +5,7 @@ import { AppContainer as HotContainer } from 'react-hot-loader'
 
 import { store, persistor } from 'configureStore'
 import MainComponent from 'components/MainComponent'
+import 'appInsights'
```
pdimitratos: This should probably change to services/appInsightsService

src/appInsights.js
```diff
@@ -0,0 +1,17 @@
+import config from 'config'
+/*
+To collect end-user usage analytics about your application, 
+insert the following script into each page you want to track.
+Place this code immediately before the closing </head> tag,
+and before any other scripts. Your first data will appear 
+automatically in just a few seconds.
+Please find more info here: https://docs.microsoft.com/en-us/azure/application-insights/app-insights-javascript
```
pdimitratos: http://apmtips.com/blog/2015/03/18/javascript-snippet-explained/ has more detailed info on this particular snippet.

This file can likely be merged into appInsightsService; if it can't, then appInsights can be exported from here instead of appInsightsService.

src/services/appInsightsService.js
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line no-undef
+export default appInsights
```
pdimitratos: The appInsights instance is likely accessible in this scope as window.appInsights. That said, it's probably best to merge this into src/appInsights.js or to merge that file into this.

src/components/Timeline/Playbook/Play.js
```diff
@@ -2,9 +2,16 @@ import React from 'react'
 import { connect } from 'react-redux'
 import FlatButtonStyled from 'components/elements/FlatButtonStyled'
 import { fillTemplate, publishEvent } from 'services/playbookService'
+import config from 'config'
```
pdimitratos: Shouldn't need config here anymore

src/components/Timeline/Playbook/Play.js
```diff
@@ -2,9 +2,16 @@ import React from 'react'
...
+import appInsights from 'src/appInsights'
+
+var trackLinkClick = (name) => {
+  // eslint-disable-next-line no-undef
```
pdimitratos: Shouldn't need the eslint directive now that appInsights is imported rather than assumed to be in the global scope

cfg/base.js
```diff
@@ -35,11 +36,11 @@ const config = {
     extensions: ['.js'],
     alias: {
       actions: `${srcPath}/actions/`,
-      components: `${srcPath}/components/`,
```
macrobug: why this change?

cfg/base.js
```diff
@@ -35,11 +36,11 @@ const config = {
...
       sources: `${srcPath}/sources/`,
       stores: `${srcPath}/stores/`,
       styles: `${srcPath}/styles/`,
       config: `${srcPath}/config/`,
+      src: `${srcPath}/`,
```
macrobug: why this change?

src/actions/authActions.js
```diff
@@ -1,16 +1,24 @@
+import config from 'config'
+//import appInsights from 'src/appInsights'
```
macrobug: remove commented out code

src/actions/authActions.js
```diff
@@ -1,16 +1,24 @@
...
 export const loginInProgress = () => ({
   type: LOGIN_IN_PROGRESS
 })
 
-export const userLoggedIn = (user) => ({
-  type: USER_LOGGED_IN,
-  user
-})
+export const userLoggedIn = (user) =>{
+  appInsights.setAuthenticatedUserContext(user.userName.replace(/[,;=| ]+/g, "_"), null, true);
```
macrobug: is this the right place for this code? consider moving to the reducer. consult with Philip what is the right approach.

src/components/Timeline/Playbook/Play.js
```diff
@@ -2,9 +2,16 @@ import React from 'react'
 import { connect } from 'react-redux'
 import FlatButtonStyled from 'components/elements/FlatButtonStyled'
 import { fillTemplate, publishEvent } from 'services/playbookService'
+import config from 'config'
+//import appInsights from 'src/appInsights'
```
macrobug: remove commented out code

src/components/Timeline/Playbook/Play.js
```diff
@@ -2,9 +2,16 @@ import React from 'react'
...
+import appInsights from '../../../appInsights'
```
macrobug: remove ../../../ relative path. this should work: 'services/appInsights'

src/components/Timeline/Playbook/Play.js
```diff
@@ -2,9 +2,16 @@ import React from 'react'
...
+
+var trackLinkClick = (name) => {
```
macrobug: since trackLinkClick never changes, use `const`

src/index.html
```diff
@@ -1,5 +1,6 @@
 <!doctype html>
 <html>
+  <script type="text/javascript" src="/assets/appInsights.js"></script>
```
macrobug: According to the appinsights docs you pasted earlier, this should go before `</head> `not before `<head>`

cfg/defaultConstants.js
```diff
@@ -39,5 +39,9 @@ module.exports = {
...
       ticketUriSuffix: '&edit=false'
     }
-  }
-}
+  },
+
+  instrumentationKey: 'APP INSIGHT GOES HERE',
+  
+  useAppInsights: true
+}
```
pdimitratos: Don't forget to end files with a line break for git compatibility.

src/index.html
yingzhao-1: Application insights works correctly by injecting the codes here. So it should be fine leaving it this way.

src/components/Timeline/Playbook/Play.js
yingzhao-1: In which file was 'services' defined? I cannot find it.

yingzhao-1: And appInsights.js is not under services.

PR #77, opened 2018-02-09
Provide timestamps in multiple time zones for events.

strand:
Connects https://azsrels.visualstudio.com/Azure%20SRE%20and%20Live%20Site/_workitems/edit/45216


src/actions/engagementActions.js
```diff
@@ -9,14 +9,14 @@ export const DISENGAGE_SUCCESS = 'DISENGAGE_SUCCESS'
 export const DISENGAGE_FAILURE = 'DISENGAGE_FAILURE'
 export const ENGAGEMENTS = 'ENGAGEMENTS'
 
-export const engage = (incidentId, participant, timeEngaged = moment()) =>
+export const engage = (incidentId, participant, timeEngaged = DateTime.local()) =>
```
pdimitratos: Engagements should be tracked only in UTC

pdimitratos: And yes, I realize that the existing code probably didn't track in UTC. Sorry about that.

src/services/playbookService.js
```diff
@@ -29,7 +29,7 @@ export const GetComparisonValue = (condition) => {
       return condition.comparisonValue
     case 1: // datetime
       return condition.dateTimeComparisonValue
-                ? moment.utc(condition.dateTimeComparisonValue)
+                ? DateTime.fromISO(condition.dateTimeComparisonValue)
```
pdimitratos: I'm going to preface this with: I'm sorry for asking you to fix my bug.

This function is trying to get an object to use in a comparison in testableTestByConditionType below.

Unfortunately, testableTestByConditionType will grab a string value as condition.value even when the condition is attempting to compare datetimes. Is checking comparisonValue for being a DateTime instance and calling DateTime.fromISO(condition.value) to make const value a datetime instance and then doing whatever approved Luxon dateTime comparison rather than the standard JavaScript (value > comparisonValue) in scope for this pull request, or should we open a separate bug for it?

src/helpers/timeFormattedToMultipleZones.js
```diff
@@ -0,0 +1,53 @@
...
+// }
+
+// dateGroup) => {
+//   return _.map(dateGroup, (datetime, date) => {
+//     debugger
+//     return _.first(dateGroup[date]) === datetime ?
+//               datetime.toLocaleString(format) :
+//               datetime.toLocaleString(DateTime.TIME_24_WITH_SECONDS)
+//   }).join(' lejoiner ')
+// }
```
pdimitratos: Comments and debugger statements should be cleaned up before we merge.

src/helpers/timeFormattedToMultipleZones.js
```diff
@@ -6,48 +6,38 @@ const zones = [
...
     return timepoint
   }
 
+  let timeAndDate = (timepoint, index) => {
+    let format = index === 0 ? timeAndDateformat : DateTime.TIME_24_WITH_SECONDS
+
+    return timepoint.timeInZone.toLocaleString(format) + ' ' + timepoint.shortname
+  }
+
+  let functionName = (dateGroup) => _.map(dateGroup, timeAndDate).join(', ')
```
strand: rename to `transformToTimeAndDateString`

src/helpers/timeFormattedToMultipleZones.js
```diff
@@ -0,0 +1,48 @@
+import { DateTime } from 'luxon'
+import _ from 'underscore'
+
+// PST https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
+// India's IANA zone https://en.wikipedia.org/wiki/Time_in_India
+const zones = [
+  { shortname: 'PT', ianaZone: 'America/Los_Angeles' },
+  { shortname: 'IST', ianaZone: 'Asia/Kolkata' },
+  { shortname: 'GMT', ianaZone: 'Etc/GMT' }
```
macrobug: Any particular reason we are calling this GMT instead of UTC? UTC seems to be the most commonly used name across Azure.

strand: Good catch.

src/services/playbookService.js
strand: I don't think I understand this bug, and it was not introduced by this change. Can you open  followon ticket with reproduction steps?

strand:
This PR transitions us from moment.js to Luxon, and there is a slight possibility that there is breaking functionality on untested code. 

It adds a transform-to-multiple-timezones function which defaults to providing the user with time in PT, IST, and UTC.

I slightly increased test coverage by deleting a dead test.

PR #78, opened 2018-02-09
SIA Chrome Extension Push Notifications

macrobug:
_This is still work in progress._ 

Seems pretty stable at this point. However, the playbook integration is somewhat experimental and may not work under some circumstances.

**Features:**
- [x] Display push notification when a new event arrives.
- [x] Display count of new notifications in the extension browser icon.
- [x] Ability to turn on and off notifications by right-clicking on the extension browser icon.
- [x] Clicking on the notification opens the SIA page.
- [x] _(Experimental)_ Ability to execute playbook actions from the notification (Chrome only supports up to two buttons).
- [x] Unit tests.


src/components/Extension/ExtensionBadge.js
```diff
@@ -0,0 +1,12 @@
+import { Component } from 'react'
+
+class ExtensionBadge extends Component {
+  render () {
+    if (chrome && chrome.browserAction) {
+      chrome.browserAction.setBadgeText({ text: this.props.text })
+    }
+    return null
```
macrobug: This is a "renderless" component. Doesn't draw anything on the page, instead uses the chrome API to update the browser icon.

src/components/Extension/Notification.js
```diff
@@ -0,0 +1,105 @@
...
+    const title = ticket.data ? ticket.data.title : null
+
+    const message = eventData.DisplayText ||
+                    Object.entries(eventData)
+                          .map(([key, value]) => `${key}: ${value}`)
+                          .join('\n')
+
+    const defaultActions = [{
+      title: 'Open in IcM', // TODO: Ticket system hardcoded
+      url: `https://icm.ad.msft.net/imp/v3/incidents/details/${ticketId}/home`
```
macrobug: Need to get these from the config.

strand:
There are a lot of changes in this commit, but many of them are focused on configuration or improving existing implementation. I would focus my time testing the code which has reduced coverage and the code which is added in this commit.

The list of files/methods which are lower hanging fruit on which we could improve test coverage of is:

- `actions/eventActions.js#getEventActionSet`
- `components/Extension/ExtensionBadge.js#render`
- `components/Extension/Notification.js#createNotification` // Note: this caught my attention because it is a long method.
- `components/Extension/Notifications.js#render`
- `reducers/notificationReducer.js#notificationReducer`

src/actions/notificationActions.js
```diff
@@ -0,0 +1,91 @@
...
+                                        chrome.extension &&
+                                        chrome.extension.getBackgroundPage)()
+
+export const isChromeExtensionBackground = (() => isChromeExtension &&
+                                                  chrome.extension.getBackgroundPage() === window)()
+
+export const isChromeExtensionPopup = (() => isChromeExtension &&
+                                             chrome.extension.getBackgroundPage() !== window)()
+
+export const registerExtensionActions = (dispatch) => {
```
pdimitratos: Generally things that are 'do-once' aren't really 'actions'; it might be best to move this to a 'notificationsService'. Compare to signalRService which imports signalRActions and registers events that dispatch those actions.

src/components/Extension/Notification.js
```diff
@@ -0,0 +1,105 @@
...
+      buttons: buttonActions.map(buttonAction => ({
+        title: buttonAction.title,
+        iconUrl: '/static/sia-green-icon-128.png'
+      }))
+    }
+
+    const msg = `############## EventId: ${this.props.eventId} -- createNotification.`
+    console.log('%c' + msg, 'color:Green;font-weight:bold;', this.props, notificationBody)
+
+    if (chrome && chrome.notifications) {
```
pdimitratos: Is there even a reason to render mount this component if this condition isn't met?

src/reducers/notificationReducer.js
```diff
@@ -0,0 +1,50 @@
...
+  receivePushNotificationsEnabled: true,
+  list: []
+}
+
+export default function notificationReducer (state = initialState, action) {
+  switch (action.type) {
+    case notificationActions.EMIT_NOTIFICATION:
+      return {
+        ...state,
+        list: state.receivePushNotificationsEnabled ? state.list.concat(action.notification) : []
```
pdimitratos: It can be dangerous for a reducer to rely too much on its own state to determine how it responds to actions. Including this dependency on receivePushNotificationsEnabled means that we can never break this reducer into multiple using combineReducers (such as to have a more complex set of notification options as a separate reducer to the list of notifications). I think it's better to avoid dispatching actions if we don't intend for them to alter the state (or to be used expressly for debugging, like rawHttp and json actions). I'd prefer to have the notification component pull receivePushNotificationsEnabled out of state and have the list reducer separate from the options reducer. This has the added benefit of significantly simplifying testing of the reducer, as there are fewer cases to account for overall.

src/components/Extension/Notification.js
```diff
@@ -0,0 +1,105 @@
...
+      eventType,
+      engagement
+    } = this.props
+
+    const ticketId = ticket.originId
+    const status = ticket.status
+    const severity = ticket.data ? ticket.data.severity : null
+    const title = ticket.data ? ticket.data.title : null
+
+    const message = eventData.DisplayText ||
```
pdimitratos: import { LoadTextFromEvent } from 'services/playbookService'

cfg/base.js
```diff
@@ -32,16 +33,7 @@ const config = {
     noInfo: false
   },
   resolve: {
-    extensions: ['.js'],
-    alias: {
```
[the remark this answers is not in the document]
macrobug: These are not being used. The aliases are defined in `.babelrc` under `module-resolver`. Forgot to remove these back then.

.gitignore
```diff
@@ -2,6 +2,8 @@
 logs
 *.log
 
+dist/*
```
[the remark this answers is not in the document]
maggiepint: The build ends up with our Microsoft endpoints in the public distro - which is not really bad (the endpoints are public) - but kinda useless for the broader community and weird.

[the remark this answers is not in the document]
macrobug: `dist` is generated by running `npm run dist`

Ideally we want to avoid having generated artifacts in the repo unless there is a good reason to. 

These generated artifacts will include stuff such as our configuration, which is not really useful for third parties forking our repo. For example, `\dist\assets\app.js`, will contain the link to IcM, clientId, etc.

src/components/Extension/Notification.js
macrobug: You are right. This was for debugging without having to install the extension. Will make the change.

src/reducers/notificationReducer.js
macrobug: Interesting, let's talk more about this when you have a minute.

src/components/Extension/Notification.js
macrobug: Taking a look...

src/actions/notificationActions.js
macrobug: Will take a look.

test/services/notificationServiceTest.js
```diff
@@ -18,56 +18,99 @@ describe('Notification Service', function () {
...
+            contextMenus: {
+              onClicked: { addListener: () => { runtimeOnMessageCalled = true } },
+              create: () => null
+            },
+            notifications: {
+              onClicked: { addListener: () => { runtimeOnMessageCalled = true } },
+              onButtonClicked: { addListener: () => { runtimeOnMessageCalled = true } },
+              onClosed: { addListener: () => { runtimeOnMessageCalled = true } } }
+          }
+          // global.chrome.runtime.onMessage.addListener
```
strand: delete comments

PR #80, opened 2018-02-14
Redraw events timeline when a link is followed

strand:
See also: https://azsrels.visualstudio.com/Azure%20SRE%20and%20Live%20Site/_workitems/edit/45579

### DIFF
--- a/src/components/Timeline/Timeline.js
+++ b/src/components/Timeline/Timeline.js
@@ -30,9 +30,10 @@ class Timeline extends Component {
   }
 
   componentDidUpdate (oldProps) {
-    const { dispatch, incidentId, ticketId } = this.props
+    const { dispatch, history, incidentId, filter, ticketId } = this.props
+
     if (oldProps.incidentId !== incidentId) {
-      dispatch(eventActions.fetchEvents({ ticketId: ticketId, incidentId: incidentId }))
+      dispatch(filterActions.synchronizeFilters(filter, incidentId, ticketId, history))
       updatePagination(incidentId, dispatch)
     }
   }
--- a/src/components/Timeline/Timeline.js
+++ b/src/components/Timeline/Timeline.js
@@ -29,17 +29,12 @@ class Timeline extends Component {
     }
   }
 
-  shouldComponentUpdate (newProps) {
-    const { eventTypes, events, ticketId, incidentId, filter, history, dispatch } = newProps
-    if (this.props.ticketId !== newProps.ticketId) {
+  componentDidUpdate (oldProps) {
+    const { eventTypes, events, ticketId, incidentId, filter, history, dispatch } = this.props
+    if (oldProps.incidentId !== incidentId) {
+      dispatch(eventActions.fetchEvents({ ticketId: ticketId, incidentId: incidentId }))
       updatePagination(incidentId, dispatch)
-      fetchMissingEventTypes(eventTypes, events, dispatch)
-      if (incidentId) {
-        dispatch(filterActions.synchronizeFilters(filter, incidentId, ticketId, history))
-      }
     }
-
-    return true
   }
 
   render () {
@@ -68,12 +63,13 @@ const fetchMissingEventTypes = (eventTypes, events, dispatch) => {
 }
 
 const mapStateToProps = (state, ownProps) => {
-  const { events } = state
+  const { events, eventTypes } = state
+
   return {
     ...ownProps,
     events: events.pages,
     filter: events.filter,
-    eventTypes: state.eventTypes.records
+    eventTypes: eventTypes.records
   }
 }
 
--- a/src/components/Timeline/Timeline.js
+++ b/src/components/Timeline/Timeline.js
@@ -29,6 +29,19 @@ class Timeline extends Component {
     }
   }
 
+  shouldComponentUpdate (newProps) {
+    const { eventTypes, events, ticketId, incidentId, filter, history, dispatch } = newProps
+    if (this.props.ticketId !== newProps.ticketId) {
+      updatePagination(incidentId, dispatch)
+      fetchMissingEventTypes(eventTypes, events, dispatch)
+      if (incidentId) {
+        dispatch(filterActions.synchronizeFilters(filter, incidentId, ticketId, history))
+      }
+    }
+
+    return true
+  }
+
   render () {
     const { events, dispatch, ticketId, incidentId, eventTypes, history } = this.props
     return (
--- a/src/components/Timeline/Timeline.js
+++ b/src/components/Timeline/Timeline.js
@@ -30,7 +30,7 @@ class Timeline extends Component {
   }
 
   componentDidUpdate (oldProps) {
-    const { eventTypes, events, ticketId, incidentId, filter, history, dispatch } = this.props
+    const { dispatch, incidentId, ticketId } = this.props
     if (oldProps.incidentId !== incidentId) {
       dispatch(eventActions.fetchEvents({ ticketId: ticketId, incidentId: incidentId }))
       updatePagination(incidentId, dispatch)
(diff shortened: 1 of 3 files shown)


src/components/TopNav/NavMenu.js
```diff
@@ -8,11 +8,15 @@ import NavigationMenu from 'material-ui/svg-icons/navigation/menu'
 import IconButton from 'material-ui/IconButton'
 import * as auth from 'services/authNService'
 
-var transformIdToTicketLink = (id, index) =>
-  <MenuItem
-    key={'ticket-' + id + '-index-' + index}
-    primaryText={<Link to={'/tickets/' + id} >{'Ticket ' + id}</Link>}
-  />
+var transformIdToTicketLink = (id, index) => {
+  let addCurrentPageToHistory = () => { history.push(this.to) /* global history */ }
```
strand: [This comment](https://github.com/Azure/Sia-EventUI/pull/75/commits/2e2b95fe457b341b781319c28cfb07bbee5bf945#r165778005) indicates we should inject this change at the Link level. Initial investigation didn't yield an obvious way to achieve this.

src/components/TopNav/NavMenu.js
```diff
@@ -35,9 +39,11 @@ NavMenu.propTypes = {
 export const mapStateToProps = (state, ownProps) => {
   var pathname = ownProps.location.pathname
   var currentId = /\d/.test(pathname) && pathname.match(/(\d+)/)[1]
+  let idContainsANumberAndIsNotCurrent = (id) => id !== currentId && /\d/.test(id)
```
strand: Adding `/\d/.test` prevents 'permit' from being listed as a ticket. https://azsrels.visualstudio.com/_api/_wit/DownloadAttachment?fileName=Screen+Shot+2018-02-14+at+10.43.34+AM.png&attachmentId=829126&contentOnly=true&__v=5

src/components/TopNav/NavMenu.js
strand: I think that the Timeline/Event components failing to load appropriate content [has something to do with blocked updates.](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/guides/blocked-updates.md)

I attempted to brute force this by wrapping most `connect()` calls in `withRouter`, but this does not resolve the issue. I think I need to pair with someone hwo has greatr React knowledge to correctly resolve this.

src/components/Timeline/Event.js
```diff
@@ -11,81 +11,89 @@ import LoadingMessage from 'components/elements/LoadingMessage'
...
-                  engagementId={engagementId}
-                />
-              </CardText>
-            }
-              </Card>
-            </div>
-}
+class Event extends Component {
+  static propTypes = {
+    text: PropTypes.string.isRequired,
```
strand: sort alphabetically.

src/components/Timeline/Event.js
```diff
@@ -11,81 +11,89 @@ import LoadingMessage from 'components/elements/LoadingMessage'
...
-  backgroundColor: PropTypes.string,
-  ticketId: PropTypes.string,
-  eventId: PropTypes.number,
-  eventTypeId: PropTypes.number,
-  eventTypeIsFetching: PropTypes.bool,
-  event: PropTypes.object
-}
+  render () {
+    const {
+            text,
```
strand: sort

src/components/Timeline/Event.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { Component } from 'react'
```
strand: Nothing in this file is necessary for this change to work, revert to master.

src/components/TopNav/NavMenu.js
```diff
@@ -8,11 +8,13 @@ import NavigationMenu from 'material-ui/svg-icons/navigation/menu'
 import IconButton from 'material-ui/IconButton'
 import * as auth from 'services/authNService'
 
-var transformIdToTicketLink = (id, index) =>
```
strand: revert to original implementation

src/components/Timeline/Timeline.js
```diff
@@ -29,6 +29,19 @@ class Timeline extends Component {
     }
   }
 
+  shouldComponentUpdate (newProps) {
```
strand: Use `componentDidUpdate` or `componentWillUpdate` instead.

src/components/Timeline/Timeline.js
```diff
@@ -29,6 +29,19 @@ class Timeline extends Component {
...
+    const { eventTypes, events, ticketId, incidentId, filter, history, dispatch } = newProps
+    if (this.props.ticketId !== newProps.ticketId) {
+      updatePagination(incidentId, dispatch)
```
strand: These actions are not giving the desired effect. The state we're losing is the events associated with the new incident… I believe what's occurring is we assume the events in the timeline will remain the same, and so when we click on the breadcrumb list it doesn't fetch new events. @pdimitratos Can we pair on this on Monday?

strand: Nevermind, I figured it out.

src/components/Timeline/Timeline.js
```diff
@@ -29,6 +29,14 @@ class Timeline extends Component {
     }
   }
 
+  componentDidUpdate (oldProps) {
+    const { dispatch, incidentId, ticketId } = this.props
+    if (oldProps.incidentId !== incidentId) {
+      dispatch(eventActions.fetchEvents({ ticketId: ticketId, incidentId: incidentId }))
```
pdimitratos: I don't think this preserves the event type and other filters.

I think this will put us in a situation where the filter component may show that only a few event types are selected, but the API calls will load all event types.

SynchronizeFilters includes a call to fetchEvents; it may be best to use that.

strand: Replacing `dispatch(eventActions.fetchEvents({ ticketId: ticketId, incidentId: incidentId }))` with `dispatch(filterActions.synchronizeFilters(filter, incidentId, ticketId, history))` results in buggy behavior. Specifically, using just synchronizeFilters causes the event timeline to be depopulated when a user clicks on a link. If I keep both lines, the behavior is correct, but I'm confused, because from what you've told me, synchronizeFilters should be implicitly fetching.

pdimitratos: We don't do client side filtering on anything but incidentId because our pagination library only supports filtering on one thing at a time; when we change filter we drop the event collection and reload it from the server. I think the bug may come from filteractions.synchronizeFilters:

    const newFilter = Object.assign({ incidentId: incidentId, ticketId: ticketId }, filter)

If the passed in filter has a different incidentId from the incidentId passed into the function, it will use the passed in (*old*) filter. That causes it to load events for the old incidentId, but UpdatePagination will cause our pagination library to filter down to only events with the new incidentId, leaving none. Since you're passing the new filter into eventActions.fetchEvents, that gets the events for the new incident, which may be why the behavior is right when you have both.

    const newFilter = Object.assign({}, filter, { incidentId: incidentId, ticketId: ticketId })

^This in synchronizeFilters may result in the correct behavior using sychronizeFilters alone.

strand: s̵͎̝̜̏̿͝u̸̪̜̞̾́̊ḇ̵̪͓̓͒̒t̷̨͓̖̀̃͂l̷̢̲̥̆̈́͠e̸̫͇̫͊̋̆ ̴̼̯̼̆̊̚b̴̢̼̳͊̑́u̸̢͎͙͐͋̽ģ̸̱̭̊̀͝s̵͔̮̝͗̓̄

strand: @pdimitratos Theory confirmed: screen shot 2018-02-21 at 3 59 43 pm

strand: I especially like the `Object.assign({},` syntax, though it is too clever by half, it does great coercing for empty values.

strand:
@pdimitratos I'm 99% certain this still does what it says on the tin, tho I will to do a local verification before this is merged… can you CR and approve?

PR #84, opened 2018-02-24
version bump

maggiepint:


maggiepint:
That’s something we should invest in. Right now, I just have to bump the version every time I cut a release of the chrome extension or it won’t accept it.

PR #85, opened 2018-02-26
[MainComponent] Load Router instead of BrowserRouter

strand:
Resolves https://azsrels.visualstudio.com/Azure%20SRE%20and%20Live%20Site/_workitems/edit/45918

### DIFF
--- a/src/components/MainComponent.js
+++ b/src/components/MainComponent.js
@@ -2,7 +2,7 @@ import React from 'react'
 import { Provider } from 'react-redux'
 import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider'
 import getMuiTheme from 'material-ui/styles/getMuiTheme'
-import { BrowserRouter as Router, Route } from 'react-router-dom'
+import { Router, Route } from 'react-router-dom'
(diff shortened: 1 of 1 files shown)


macrobug:
Should this require a unit test? Were we using any feature from BrowseRouter at all?

strand:
@maggiepint Is removing the BrowserRouter dependency an acceptable solution here?

strand:
@macrobug If I had any idea what behavior changes this will cause, if any, I would have written a spec.

macrobug:
@strand I have no idea neither. If routing seems to work fine, as in one can navigate across pages, then it should be ok.

maggiepint:
I believe that when I added browser router back in the day, it was because that preserved the history with the back button. I think that react router has revved major versions since then, and it may be that this code is just plain deprecated. If the back button still works, then I imagine this PR is fine.

strand:
There is a back button bug which exists on master without this change. When navigating the application, the back button requires two presses to go back one page, another three (for a total of five) to go back two pages, and another three (total of 8) clicks to go back three pages.

back-button-bug

However, since this bug exists in _either_ master or this PR, I think this bug is not related to this work.

PR #86, opened 2018-02-27
added docker build

zhz2017:
1. Docker files to build localhost and dist images.
2. A modified createconstants.ps1 is put in deployment directory. It use combination of environment variables and input parameters to generate configuration consts for localhost and dist environments. Existing createconstants.ps1 in cfg directory will be removed whence we have the build definition worked out.
3. Fixed an issue in MainComponents.js so it works with Linux container which a file name is case sensitive on letters.


deployment/createconstants.ps1
```diff
@@ -0,0 +1,46 @@
+# Powershell script for VSTS build that accesses build environment variables to create config file.
```
pdimitratos: Does this replace the other createconstants.ps1? Do we have a reason to maintain two versions if it doesn't?

zhz2017: Yes. Whence we have EventUI Build working, cfg/createconstants.ps1 will be removed.

PR #88, opened 2018-02-27
[WIP] Create and extract checklist plugin

strand:


src/components/MainComponent.js
```diff
@@ -17,6 +17,26 @@ import Debug from 'components/Debug'
 import { isChromeExtensionBackground } from 'services/notificationService'
 import Notifications from 'components/Extension/Notifications'
 
+try {
+  const Checklist = require('components/Checklist')
+}
+catch(error) {
+  if ( !error.toString().includes('Cannot find module') ) {
```
macrobug: I believe you can access the error code here:
`if (error.code && error.code === 'MODULE_NOT_FOUND') {`

src/components/MainComponent.js
```diff
@@ -17,6 +17,26 @@ import Debug from 'components/Debug'
...
+    throw error
+  } else {
+
+  }
+}
+
+const isPluginDefined = () => {
+  if ( typeof Checklist !== undefined ) {
+    return false
```
macrobug: I think the conditional is backwards. Should return true if not undefined:
`const isPluginDefined = () => typeof Checklist !== undefined`

src/components/MainComponent.js
```diff
@@ -17,6 +17,26 @@ import Debug from 'components/Debug'
...
+}
+catch(error) {
+  if ( !error.toString().includes('Cannot find module') ) {
+    throw error
+  } else {
+
+  }
+}
+
+const isPluginDefined = () => {
```
pdimitratos: This should also work as 

    const isPluginDefined = () => (typeof Checklist !== undefined)

Actually, I'm now slightly confused, because the above line returns the opposite of the current version. isPluginDefined returns true when Checklist is undefined?

loadPluginsToTempDir.js
```diff
@@ -0,0 +1,55 @@
...
+
+  plugin.newPath = [directoryPath, filename].join(separator)
+
+  return plugin
+}
+
+const plugins = (environment = process.env) => {
+  if ( !environment.PLUGINS ) { console.log('No plugins are defined.'); return }
+
+  return _.map(JSON.parse(environment.PLUGINS), hydratePlugin)
```
pdimitratos: Does it make sense to have an environment.PLUGINS.example.js file somewhere to show what this configuration should look like in a way that can be checked into source control?

src/components/MainComponent.js
```diff
@@ -31,6 +51,7 @@ export default class MainComponent extends React.Component {
                   <div>
                     { isChromeExtensionBackground() ? <Notifications /> : null }
                     <TopNav />
+                    { isPluginDefined() ? <Route path='/tickets/:ticketId/checklist' component={Checklist} /> : null}
```
pdimitratos: Seems like it could be useful to have a PluginRoute component that takes like path, component (component name to use) and pluginPath (for the require statement). That could also standardize the user experience when a component isn't available; I'd rather see an error screen saying something like 'Checklist plugin could not be loaded, please contact your system administrator' than just not being able to navigate to a route (which may have been working the previous day prior to some breaking update).

loadPluginsToTempDir.js
```diff
@@ -0,0 +1,55 @@
...
+  _.each(plugins(), copyFile)
+}
+
+const teardown = () => {
+  // Not triggered by ctrl-C. :(
+  console.log('Removing temp plugin folder.')
+  fs.rmdir(directoryPath, handleError)
+}
+
+if (process.argv.includes('--setup')) {
```
pdimitratos: Call teardown() here before setup()? That would remove any 'pollution' of the directory by previous setup that wasn't accompanied by a teardown (such as with ctrl-c on the node cli or (I assume) power loss to the node.

src/components/MainComponent.js
```diff
@@ -17,6 +17,48 @@ import Debug from 'components/Debug'
...
+
+//
+
+// _.each(plugins(), function (plugin) {
+//
+// })
+
+const maybePlugin = () => {
+  try {
+    const Checklist = require('components/Checklist') // eslint-disable-line no-unused-vars
```
pdimitratos: Does a try { } block create a new scope? I would have thought that this would need to be like

    let Checklist
    try {
       Checklist = require()
    }

PR #91, opened 2018-02-28
generating const configs with bash script

zhz2017:
We found environment variables are not passed in powershell task when using VSTS Linux agent. The PR replaces deployment/createconstants.ps1 with CreateConstants.sh bash script to generate const configs.


zhz2017:
Only modified createconstants.ps1 in deployment directory is removed. The original createconstants.ps1 is still in cfg directory for fallback. Windows agent supports Bash task too. However, I am fine with keeping the copy in cfg directory after the Build working just in case we can use it in some ways.

PR #92, opened 2018-03-01
importing a modular component from github

strand:


strand:
I’m definitely thinking about making this more generic and configurable, but for now I’m gonna follow PEP-20 and favor the explicit over the implicit.

PR #93, opened 2018-03-02
Added more tests, attempted to add coverage yaml config for categoriz…

pdimitratos:
…ation of coverage


src/actions/actionHelpers.js
```diff
@@ -1,33 +1,44 @@
...
-export const reduxBackedPromise = (promiseArgs, actionSet, operation = 'GET') => (dispatch) => {
-  if (!actionSet.try) { throw needOnActionSet('try') }
-  if (!actionSet.succeed) { throw needOnActionSet('succeed') }
-  if (!actionSet.fail) { throw needOnActionSet('fail') }
+export const validateActionSet = (actionSet) => {
+  let need
+  if (!actionSet.fail || !(typeof actionSet.fail === 'function')) { need = 'fail' }
+  if (!actionSet.succeed || !(typeof actionSet.succeed === 'function')) { need = 'succeed' }
+  if (!actionSet.try || !(typeof actionSet.try === 'function')) { need = 'try' }
+  if (need) { throw needOnActionSet(need) }
```
macrobug: throwing an Error instead of string is usually preferable:
  `throw new Error(needOnActionSet(need))`

Depending on the browser, it may include additional contextual information, and has better cross-browser compatibility.

src/actions/actionHelpers.js
```diff
@@ -1,33 +1,44 @@
...
+export const reduxBackedPromise = testableReduxBackedPromise(authenticatedFetch, authenticatedPost, authenticatedPut)
+
 const needOnActionSet = (prop) => `Need "${prop}" function on actionSet!`
 
-export const reduxBackedPromise = (promiseArgs, actionSet, operation = 'GET') => (dispatch) => {
-  if (!actionSet.try) { throw needOnActionSet('try') }
-  if (!actionSet.succeed) { throw needOnActionSet('succeed') }
-  if (!actionSet.fail) { throw needOnActionSet('fail') }
+export const validateActionSet = (actionSet) => {
+  let need
```
macrobug: The above can by simplified:

```
const need = ['fail', 'succeed', 'try'].filter(a => !actionSet[a] || typeof actionSet[a] !== 'function')
if (need.length) { 
  throw new Error(need.map(needOnActionSet).join())
}
```

PR #96, opened 2018-03-05
Yizha/load uncorrelated events

yingzhao-1:


src/actions/filterActions.js
```diff
@@ -3,6 +3,7 @@ import ByPath from 'object-path'
 
 import * as eventActions from 'actions/eventActions'
 import * as filterService from 'services/filterService'
+import moment from 'moment'
```
macrobug: I believe moment was replaced with luxon.

src/actions/filterActions.js
```diff
@@ -14,6 +15,21 @@ export const changeEventFilter = (history) => (filter) => {
   }
 }
 
+export const changeUncorrelatedEventFilter = (history) => (filter) => {
```
pdimitratos: You can merge this with changeEventFilter by changing the signature of changeEventFilter from (history) => (filter) => {...} to (history, urlLoader = filterService.getUrlFromFilter) => (filter) => {...}

Since the only difference between changeEventFilter and changeUncorrelatedEventFilter is:

filterService.getUrlFromFilter(history, filter)
vs
filterService.getUrlFromUncorrelatedFilter(history, filter)

That line can become

urlLoader (history, filter)

Then the call to changeUncorrelatedEventFilter becomes

dispatch(changeEventFilter(history, filterActions.getUrlFromUncorrelatedFilter)(newFilter))

src/actions/filterActions.js
```diff
@@ -14,6 +15,21 @@ export const changeEventFilter = (history) => (filter) => {
...
+  filterService.getUrlFromUncorrelatedFilter(history, filter)
+  return {
+    type: CHANGE_EVENT_FILTER,
+    filter
+  }
+}
+
+export const clearFilterIncidentId = (filter) => {
```
pdimitratos: This doesn't actually clear the Filter Incident Id (though I see you do so manually before calling this function).

The fact that this (an action that specifically controls the incidentId value of the filter) is needed makes me think that the filter reducer would be better off broken up into multiple separate reducers with a combineReducers statement; that would give us more granular control of what gets updated by each action type.

If you'd rather not go that far, you can replace the call to this action in uncorrelatedEvents.js with dispatch(filterActions.synchronizeFilters(filter, null, null, history)), though you'd also need a WithRouter() statement to pull in history.

If you definitely want to keep this action, change the type to something that matches the function name (CLEAR_EVENT_FILTER_INCIDENTID or similar) and add a reducer case for it:

const { incidentId, ...newState } = state
return newState

src/actions/filterActions.js
```diff
@@ -44,12 +60,17 @@ export const removeFilter = (history, relativeFilterPath) => (oldFilter, filterT
...
-    dispatch(changeEventFilter(history)(newFilter))
-    dispatch(eventActions.fetchEvents(newFilter))
+  if (newFilter.incidentId) {
+    if (!deepEquals(oldFilter, newFilter)) {
+      dispatch(changeEventFilter(history)(newFilter))
+      dispatch(eventActions.fetchEvents(newFilter))
+    }
+  } else {
+    const endTime = newFilter.endTime ? DateTime.fromJSDate(new Date(newFilter.endTime)) : DateTime.local()
+    const startTime = newFilter.startTime ? DateTime.fromJSDate(new Date(newFilter.startTime)) : DateTime.local().minus({days: 1})
```
pdimitratos: It may be better not to tie the hardcoded start and end times in a filter to the lack of incidentId on the filter; moving these calculations to the componentDidMount method of uncorrelatedEvents.js and including them in the filter passed into the dispatch from that method would give the same effect, but it would put the code for determining the date range shown by default on the uncorrelatedEvents component in the same file as that component, which seems preferable.

src/components/MainComponent.js
```diff
@@ -5,7 +5,6 @@ import getMuiTheme from 'material-ui/styles/getMuiTheme'
 import { Router, Route } from 'react-router-dom'
 import createBrowserHistory from 'history/createBrowserHistory'
 import { PersistGate } from 'redux-persist/lib/integration/react'
-
```
pdimitratos: This blank line between external imports and internal imports is intentional.

src/components/MainComponent.js
```diff
@@ -14,9 +13,9 @@ import incidentRedirect from 'components/Incident/IncidentRedirect'
 import Home from 'components/Home'
 import TopNav from 'components/TopNav/TopNav'
 import Debug from 'components/Debug'
+import uncorrelatedEvents from 'components/uncorrelatedEvents'
 import { isChromeExtensionBackground } from 'services/notificationService'
 import Notifications from 'components/Extension/Notifications'
-
```
pdimitratos: This blank line between imports and code is intentional.

src/components/TopNav/NavMenu.js
```diff
@@ -24,6 +24,7 @@ export const NavMenu = ({ dispatch, history, ticketIds }) => {
     <MenuItem key='logout' primaryText={<Link to='/' onClick={() => dispatch(auth.logOut)}>LogOut</Link>} />
     { ticketIds && ticketIds.map(transformIdToTicketLink) }
     <MenuItem key='debug' primaryText={<Link to='/debug' >Debug</Link>} />
+    <MenuItem key='load uncorrelated events' primaryText={<Link to='/events'>Load Events</Link>} />
```
pdimitratos: For a new end-user, this menu option may be a bit confusing. Something along the lines of 'Events For All Incidents' would clarify the difference between this page and the page for specific incident timelines.

src/components/uncorrelatedEvents.js
```diff
@@ -0,0 +1,94 @@
...
+import EventFilter from 'components/Timeline/EventFilter'
+import Footer from 'components/Timeline/EventFooter'
+import Events from 'components/Timeline/Events'
+import { DateTime } from 'luxon'
+import * as filterActions from 'actions/filterActions'
+import TimePicker from 'material-ui/TimePicker'
+import DatePicker from 'material-ui/DatePicker'
+import { FlatButton } from 'material-ui'
+
+class uncorrelatedEvents extends Component {
```
pdimitratos: By convention, component class and function names should be capitalized. See: https://reactjs.org/docs/jsx-in-depth.html#specifying-the-react-element-type

src/components/uncorrelatedEvents.js
```diff
@@ -0,0 +1,94 @@
...
+  constructor (props) {
+    super(props)
+    this.state = {startDate: null, startTime: null, endDate: null, endTime: null}
+  }
+
+  render () {
+    const { events, dispatch, eventTypes, history } = this.props
+    return (
+      <div >
+        <DatePicker
```
pdimitratos: This filtering interface looks quite useful, to the point where I would prefer to include it on the main Timeline component as well. Extracting this to its own component would allow re-use and also vastly simplify the uncorrelatedEvents component by extracting all the stuff that deals with React local state.

That said, this is another place where I think it makes a lot of sense to break the filter Reducer into multiple combined reducers; these pickers could dispatch actions to update just their portion of the Redux state. Not sure if that's worth the extra time investment for now, though.

src/actions/filterActions.js
```diff
@@ -5,15 +5,23 @@ import * as eventActions from 'actions/eventActions'
 import * as filterService from 'services/filterService'
 
 export const CHANGE_EVENT_FILTER = 'CHANGE_EVENT_FILTER'
+export const CLEAR_EVENT_FILTER_INCIDENTID = 'CLEAR_EVENT_FILTER_INCIDENTID'
 
-export const changeEventFilter = (history) => (filter) => {
-  filterService.getUrlFromFilter(history, filter)
+export const changeEventFilter = (history, urlLoader) => (filter) => {
```
pdimitratos: This seems to be breaking the build/tests because the second argument isn't passed everywhere this method is called; if you default it to filterService.getUrlFromFilter, that should fix it. Alternatively, you could change the tests to pass in the getUrlFromFilter (or other url creation function).

src/reducers/filterReducer.js
```diff
@@ -2,6 +2,9 @@ import * as filterActions from 'actions/filterActions'
 
 export const filter = (defaultFilter) => (state = defaultFilter, action) => {
   switch (action.type) {
+    case filterActions.CLEAR_EVENT_FILTER_INCIDENTID:
+      const {incidentId, ...newFilter} = action.filter
+      return {...state, ...newFilter}
```
pdimitratos: Since incidentId is already part of state, this will just add incidentId back in. You can just do:

return newFilter

yingzhao-1: no it won't, this incidentId is a property of filter, not the incidentId in the state level.

src/components/MainComponent.js
```diff
@@ -3,9 +3,9 @@ import { Provider } from 'react-redux'
 import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider'
 import getMuiTheme from 'material-ui/styles/getMuiTheme'
 import { Router, Route } from 'react-router-dom'
-import createBrowserHistory from 'history/createBrowserHistory'
 import { PersistGate } from 'redux-persist/lib/integration/react'
 
+import createBrowserHistory from 'history/createBrowserHistory'
```
pdimitratos: I think this is a third party library, not sure why it was moved to the first party imports section.

src/components/elements/TimePicker.js
```diff
@@ -0,0 +1,83 @@
...
+class TimeAndDatePicker extends Component {
+  static propTypes = {
+    dispatch: PropTypes.func.isRequired,
+    filters: PropTypes.object,
+    history: PropTypes.object.isRequired
+  }
+
+  constructor () {
+    super()
+    this.state = { startDate: null, startTime: null, endDate: null, endTime: null }
```
pdimitratos: One thing that might be useful as a quality of life improvement would be to set defaults (such as defaulting to the last 24 hours) so that people don't necessarily have to choose all 4 values before submitting.

PR #97, opened 2018-03-05
Linkchips

RedSquirrelious:
This isn't quite working yet.  It removes the ticket from the recently visisted section in the hamburger menu.  But redux-persist does its job and keeps it after refresh.


src/reducers/ticketReducers.js
```diff
@@ -45,6 +46,12 @@ const addIncidentsToState = (state, incidents) => {
   return newState
 }
 
+const removeTicketFromState = (state, ticketId) => {
+  let newState = { ...state }
+  let updatedState = _.omit(newState, ticketId)
```
macrobug: No need to add a new dependency just for this one method. This ES6 method should do the trick:
```
let newState = { ...state }
Reflect.deleteProperty(newState, ticketId)
return newState
```

macrobug: This should work too:
```
const { [ticketId]: toRemove, ...newState } = state
return newState
```

RedSquirrelious: I'd looked at that one before and it seems like a good move.  Thanks!

src/components/elements/MenuLink.js
```diff
@@ -12,7 +13,13 @@ const MenuLink = (type, id, onClick, dispatch) => {
       key={`${type}-${id}` }
       primaryText={<Link to={typeRoute} >{`${typeTitle} ${id}`}</Link>}
       rightIcon={<ActionDelete
-        onClick={() => dispatch(onClick(id)) } />}
+        onClick={
+          () => {
+            dispatch(onClick(id))
+            persistor.purge()
```
macrobug: I'm not sure the MenuLink component is the best place for this logic. Maybe move to the reducer or action?

PR #98, opened 2018-03-06
proof-of-concept of mixing-in plugins at install

strand:
This does leave the package-lock.json altered, but it should provide a reasonable hook with which to inject secret plugins.


strand:
If this direction looks ok, then I'll extract this script into a module.

bunches.js
```diff
@@ -0,0 +1,29 @@
+'use strict'
```
strand: Extract into module.

.gitignore
```diff
@@ -2,6 +2,8 @@
 logs
 *.log
 
+.env
```
strand: Document bunches install steps.

package.json
```diff
@@ -7,6 +7,8 @@
   "scripts": {
     "clean": "rimraf dist/*",
     "dist": "npm run clean && cross-env NODE_ENV=dist webpack --env=dist --progress --profile --colors --display-error-details --display-reasons --debug",
+    "preinstall": "node bunches.js --mixin-secret-plugins",
```
strand: Document bunches install steps.

.gitignore
strand: [Resolved](https://github.com/strand/bunches/pull/3/files#diff-04c6e90faac2675aa89e2176d2eec7d8R5)

PR #100, opened 2018-03-09
update copy-webpack-plugin to remove vuln dependency

strand:
Resolves ssri 5.0.0 vulnerability by upgrading it to 5.2.4

screen shot 2018-03-09 at 10 13 11 am


strand:
Verified that the app continues to load after this change.

PR #102, opened 2018-03-12
Update uglifyjs-webpack-plugin

strand:
Related to https://github.com/Azure/Sia-EventUI/pull/100


strand:
Verified that the app continues to boot up and load correctly in the browser.

PR #103, opened 2018-03-12
clear tickets from hamburger menu

RedSquirrelious:
Required a workaround for persist-redux (setting ticket data to null), but it doesn't seem to hurt anything in practice.  (revisiting a page rehydrates the ticket values in the persistor)


src/reducers/ticketReducers.js
```diff
@@ -45,6 +45,18 @@ const addIncidentsToState = (state, incidents) => {
   return newState
 }
 
+const removeTicketFromState = (state, ticketId) => {
+  let newState = { ...state }
+  newState[ticketId] = null
+  return newState
+}
+
+const removeAllTicketsFromState = (state, ticketIds) => {
```
pdimitratos: It seems strange that this action and method are named 'removeAll' when they remove a given range rather than all the tickets.

return Object.Assign({}, ...Object.keys(state).map(key => ({key: null})) )

I think this (or something similar) will remove all tickets from state. For a simpler change, try using Object.keys(state) as ticketIds rather than passing it in.

test/reducers/ticketReducersTest.js
```diff
@@ -95,6 +105,17 @@ describe('Ticket Reducers', function test () {
       expect(Object.values(this.OnReceiveIncidentFromEmpty).filter(function (ticket) { return ticket.originId === 38808134 })[0]).to.exist
       expect(Object.values(this.OnReceiveIncidentFromPopulated).filter(function (ticket) { return ticket.originId === 38808134 })[0]).to.exist
     })
+
+    it('Should replace a given ticket with null values when REMOVE_TICKET is dispatched', function() {
+      const result = map(populatedTicketList, removeTicket(38502026))
+      expect(result[38502026]).to.be.null
```
pdimitratos: Another condition that would be helpful here is 

expect(result[44444444]).to.not.be.null 

to ensure that only the intended ticket is removed

src/components/TopNav/NavMenu.js
```diff
@@ -40,7 +43,10 @@ export const mapStateToProps = (state, ownProps) => {
 
   return {
     ...ownProps,
-    ticketIds: Object.keys(state.tickets.map).filter(idContainsANumberAndIsNotCurrent)
+    ticketIds: Object.entries(state.tickets.map)
```
maggiepint: So, this means we display tickets that have a non-null second position in the array? Where is this array from?

src/components/elements/MenuLink.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
+import { Link } from 'react-router-dom'
+import MenuItem from 'material-ui/MenuItem'
+import ActionDelete from 'material-ui/svg-icons/action/delete'
+
+export const MenuLink = (type, id, onClick, dispatch) => {
+  validateType(type)
```
maggiepint: Couldn't this be done with props? I think there's a way to write custom props to handle this sort of thing, and that will work a little nicer in the error stack.

src/components/elements/MenuLink.js
```diff
@@ -0,0 +1,35 @@
...
+  var typeTitle = makeTitle(type)
```
maggiepint: Inline these calls?

src/reducers/ticketReducers.js
```diff
@@ -45,6 +45,18 @@ const addIncidentsToState = (state, incidents) => {
   return newState
 }
 
+const removeTicketFromState = (state, ticketId) => {
```
maggiepint: The kids these days seem to do it differently, but back in the day we would have put these at the bottom of the file as function expressions. I think what I really want is a separation between things that are public (exported) and things that are not for readability.

src/components/TopNav/NavMenu.js
pdimitratos: The filter statement is operating on the key-value pairs returned by Object.entries; the second position in the key-value pair is the ticket object, which is updated to null when the menu button for removing a ticket is clicked. This is a workaround because redux-persist does not track removal of keys, only updates to new non-undefined values.

src/components/elements/MenuLink.js
RedSquirrelious: The way the title portion is written requires that there be at least 2 characters.

src/reducers/ticketReducers.js
RedSquirrelious: I moved all the non-public items to the top.  They needed to be declared before they were used.

src/components/elements/MenuLink.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Link } from 'react-router-dom'
+import MenuItem from 'material-ui/MenuItem'
+import ActionDelete from 'material-ui/svg-icons/action/delete'
+
+export const MenuLink = (type, id, onClick, dispatch) => {
+  validateType(type)
+  var typeTitle = type.charAt(0).toUpperCase() + type.slice(1)
```
macrobug: This will fail if the title is less than two characters. Also, may break on RTL or other non-English setups. Try using CSS: `text-transform: capitalize`

src/reducers/ticketReducers.js
```diff
@@ -45,6 +45,28 @@ const addIncidentsToState = (state, incidents) => {
...
+  let newState = { ...state }
+  newState[ticketId] = null
+  return newState
+}
+
+const removeAllTicketsFromState = (state) => {
+  let newState = { ...state }
+  Object.keys(state).map(ticketId => { newState[ticketId] = null })
```
macrobug: why not just return an empty array []?

src/reducers/ticketReducers.js
```diff
@@ -53,6 +75,10 @@ export const map = (state = defaultTicketList, action) => {
     case CREATE_INCIDENT_SUCCESS:
     case RECEIVE_INCIDENT:
       return addIncidentToState(state, action.incident)
+    case REMOVE_TICKET:
+      return removeTicketFromState(state, action.id)
+    case REMOVE_ALL_TICKETS:
+      return removeAllTicketsFromState(state, action.ids)
```
macrobug: `return []`

src/reducers/ticketReducers.js
```diff
@@ -53,6 +75,10 @@ export const map = (state = defaultTicketList, action) => {
     case CREATE_INCIDENT_SUCCESS:
     case RECEIVE_INCIDENT:
       return addIncidentToState(state, action.incident)
+    case REMOVE_TICKET:
+      return removeTicketFromState(state, action.id)
```
macrobug: `return state.filter(ticketId=> ticketId!== action.id)`

test/reducers/ticketReducersTest.js
```diff
@@ -95,6 +104,19 @@ describe('Ticket Reducers', function test () {
       expect(Object.values(this.OnReceiveIncidentFromEmpty).filter(function (ticket) { return ticket.originId === 38808134 })[0]).to.exist
       expect(Object.values(this.OnReceiveIncidentFromPopulated).filter(function (ticket) { return ticket.originId === 38808134 })[0]).to.exist
     })
+
+    it('Should replace a given ticket with null values when REMOVE_TICKET is dispatched', function() {
+      const result = map(populatedTicketList, removeTicketFromRecent(38502026))
+      expect(result[38502026]).to.be.null
```
macrobug: This just set the ticket to null, but didn't remove it from the array. If the array is enumerated, it will still contain the undesired key.

src/reducers/ticketReducers.js
pdimitratos: The state is an object rather than an array, and redux-persist does not persist changes that involve the removal of a key, so returning an empty object would cause redux-persist to re-add all the tickets removed the next time the browser is launched.

src/reducers/ticketReducers.js
pdimitratos: Redux-persist doesn't play nice with doing it that way.

test/reducers/ticketReducersTest.js
pdimitratos: Intentional, as removal of the key entirely is not tracked in redux-persist.

src/reducers/ticketReducers.js
macrobug: got it

src/components/elements/MenuLink.js
RedSquirrelious: Good point.  Deleted the capitalization.  And removed the validation fx per Maggie's request.  Validation is with PropTypes now.

PR #104, opened 2018-03-14
eventType filter and time filter working together correctly

yingzhao-1:


src/components/Timeline/Timeline.js
```diff
@@ -31,7 +31,6 @@ class Timeline extends Component {
 
   componentDidUpdate (oldProps) {
     const { dispatch, history, incidentId, filter, ticketId } = this.props
-
```
strand: Remove whitespace-only changes.

src/components/elements/TimePicker.js
```diff
@@ -17,7 +18,26 @@ class TimeAndDatePicker extends Component {
 
   constructor () {
     super()
-    this.state = { startDate: null, startTime: null, endDate: null, endTime: null }
+    this.state = {
+      endDate: DateTime.local().toISODate().toString(),
+      endTime: DateTime.local().toISOTime().toString(),
+      startDate: DateTime.local().minus({day: 1}).toISODate().toString(),
```
strand: Instead of seperating out the date and time, pass a datetime string…

src/components/elements/TimePicker.js
```diff
@@ -17,7 +18,26 @@ class TimeAndDatePicker extends Component {
...
+      startTime: DateTime.local().toISOTime().toString()
+    }
+  }
+
+  componentDidMount () {
+    const { filters, dispatch, history } = this.props
+    const { startDate, startTime, endDate, endTime } = this.state
+    const newFilter = Object.assign({}, filters, {startTime: startDate + 'T' + startTime, endTime: endDate + 'T' + endTime})
```
strand: …which you can inline here as `startTime: startDateTime`.

src/components/elements/TimePicker.js
```diff
@@ -17,7 +18,26 @@ class TimeAndDatePicker extends Component {
 
   constructor () {
     super()
-    this.state = { startDate: null, startTime: null, endDate: null, endTime: null }
+    this.state = {
+      endDate: DateTime.local().toISODate().toString(),
```
strand: I _think_ we should use `utc` unless the local zone is important for some reason.

src/components/elements/TimePicker.js
```diff
@@ -17,7 +18,26 @@ class TimeAndDatePicker extends Component {
...
+  componentDidMount () {
+    const { filters, dispatch, history } = this.props
+    const { startDate, startTime, endDate, endTime } = this.state
+    const newFilter = Object.assign({}, filters, {startTime: startDate + 'T' + startTime, endTime: endDate + 'T' + endTime})
+    dispatch(filterActions.synchronizeFilters(newFilter, null, null, history))
+  }
+
+  componentDidUpdate (prevProps) {
+    const { filters, dispatch, history } = this.props
+    if (!deepEqual(prevProps.filters, filters)) {
```
strand: What cases does `deepEqual` cover that `===` would not?

yingzhao-1: that's a good question, I don't know.

strand: I _think_ we can avoid adding a dependency and use the same value equality function [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) here instead of deep equals…

strand: Nope, stick with `deepEquals`

screen shot 2018-03-14 at 1 35 47 pm

src/components/elements/TimePicker.js
strand: Assign and reuse a `now` variable so each time is relative to the same timepoint. (something like `let now = DateTime.utc()`

yingzhao-1: Sure.

src/components/elements/TimePicker.js
pdimitratos: So I'm not sure if different but identical objects can ever arrive here, but there's nothing explicitly preventing it; depending on the cost of deepEqual, it may make sense to check first whether the objects are the same and then do the deep equality comparison only if they're not actually the same object.

Then again, deepEqual may do that optimization internally anyway.

strand: Can we assume that other people's code is optimized until we encounter performance issues? (not because their code is/will be performant, but because premature optimization is the root of all evil.)

src/components/uncorrelatedEvents.js
```diff
@@ -18,7 +18,7 @@ class UncorrelatedEvents extends Component {
 
   componentDidMount () {
     const { filters, dispatch } = this.props
-    const newFilter = Object.assign({}, filters, {incidentId: null})
+    const newFilter = Object.assign({}, filters, { incidentId: null })
     dispatch(filterActions.clearFilterIncidentId(newFilter))
```
pdimitratos: Shouldn't need the line with

const newFilter

and shouldn't need to pass anything into clearFilterIncidentId

test/actions/FilterActionsTest.js
```diff
@@ -23,9 +23,8 @@ describe('FilterActions', function () {
   describe('clearFilterIncidentId', function(){
     const filter = {incidentId: 1, eventTypes: [1, 2]}
     const result = filterActions.clearFilterIncidentId(filter)
-    const expectedValue = {type: 'CLEAR_EVENT_FILTER_INCIDENTID', filter: {incidentId: 1, eventTypes: [1, 2]}}
     it('should return an object with a type and a filter', function(){
```
pdimitratos: Let's keep the 'should' statements in line with what's actually being checked. In this case, that might look like 'should return an object with no incidentId value'.

yingzhao-1: The FilterActionTest will not return a cleared filter, it will only return an action with the type CLEAR_eVENT_FILTER_INCIDENT.

pdimitratos: Ah, right, in that case it should return an object with the correct type and no other properties.

PR #108, opened 2018-03-20
Yizha/text filter

yingzhao-1:


test/services/filterServiceTest.js
```diff
@@ -95,7 +95,7 @@ describe('FilterService', function () {
     const filterWithEventTypesAndStartEndTimes = {startTime: 'startTime', endTime: 'endTime', eventTypes: [1, 2]}
     const filterWithNoEventTypes = {startTime: 'startTime', endTime: 'endTime'}
 
-    const expectedResult = '/events/?eventTypes=1&eventTypes=2&startTime=startTime&endTime=endTime'
+    const expectedResult = '/events/?startTime=startTime&endTime=endTime&eventTypes=1&eventTypes=2'
```
strand: Rather than changing the order of the expected result string, change your expectations to not require order-dependancy by using `include`. The expected result could be an array of parameter strings

``` javascript
const expectedResult =  [ 'eventTypes=1', 'eventTypes=2', 'startTime=startTime', 'endTime=endTime' ]
```

Then alter the expectation to read something like:

``` javascript
expect(history[0]).to.include(expectedResult)
```

src/components/Timeline/Playbook/TextFilter.js
```diff
@@ -0,0 +1,62 @@
...
+    const { filters, dispatch, history } = this.props
+    const newFilter = Object.assign({}, filters, { DataSearch: this.state.value })
+    dispatch(filterActions.synchronizeFilters(newFilter, null, null, history))
+  }
+}
+
+export const mapStateToProps = (state, ownProps) => {
+  return {
+    ...ownProps,
+    dispatch: state.dispatch,
```
pdimitratos: Dispatch is added automatically, it doesn't have to be added manually in mapStateToProps

src/components/Timeline/Playbook/TextFilter.js
```diff
@@ -0,0 +1,62 @@
...
+        <TextField
+          hintText='Search'
+          id='search-field-value'
+          value={this.state.value}
+          onChange={(event) => this.handleChange(event)}
+                />
+        <FlatButton
+          label='Search'
+          primary
+          onClick={() => this.submitSearch()}
```
pdimitratos: This component could be stateless if we separated the filter value update actions from the event fetch and url update actions; the search-field-value textfield could dispatch an action that updates the DataSearch portion of the filter reducer, and the submitSearch() onClick could be replaced with a different action that updates the history and starts the event fetch. I would go so far as to say that the filter submit button shouldn't be part of the same component as the TextFilter component or the TimeAndDatePicker component; it should be a sibling to those components. As it stands, datetime and text filters are submitted separately, which may not be intuitive to users, who may expect a workflow where they set all desired form values and then submit them all at once.

At this point, it's clear that the Filter reducer needs to be broken up into several reducers with a combineReducer statement.

test/services/filterServiceTest.js
yingzhao-1: include only works between arrays. history[0] is a string

pdimitratos: [String also has include](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes), and the [Chai api supports it](http://www.chaijs.com/api/bdd/#method_include)

```javascript
const expectedResult =  [ 'eventTypes=1', 'eventTypes=2', 'startTime=startTime', 'endTime=endTime' ]

expectedResult.forEach(expectedToken => expect(history[0]).to.include(expectedToken))
```

Could work.

src/reducers/filterReducer.js
```diff
@@ -1,15 +1,83 @@
...
+    case filterActions.CHANGE_EVENT_FILTER:
+      if (action.filter.ticketId) {
+        return action.filter.ticketId
+      } else return state
+    default:
+      return state
+  }
+}
+
+export const filter = (defaultFilter) => {
```
macrobug: Instead of checking if `defaultFilter` is null or undefined in each reducer to combine, you could set a default at the function signature, something like this:
```
const initialDefaultFilter = { something: something }
export const filter = (defaultFilter = initialDefaultFilter) => {
```

Then you could just do this in each reducer to combine:
```
ticketId: ticketId(defaultFilter.ticketId),
```

pdimitratos: export const filter = (defaultFilter = {}) => {

should be good enough

src/reducers/filterReducer.js
```diff
@@ -1,15 +1,83 @@
...
+      } else return state
+    default:
+      return state
+  }
+}
+
+export const dataSearch = (defaultDataSearch) => (state = defaultDataSearch, action) => {
+  switch (action.type) {
+    case filterActions.CHANGE_EVENT_FILTER:
+      if (action.filter.DataSearch) {
```
pdimitratos: We want to be able to clear DataSearch to remove the filter and return to the (relatively) unfiltered view. This check prevents that. I'm not sure whether this check is necessary in any of the reducers. Checking for strict equality against state may be a better way to ensure we're changing the state when we want to and not changing it otherwise.

```JavaScript
case filterActions.CHANGE_EVENT_FILTER: 
  return action.filter.DataSearch === state
    ? state
    : action.filter.DataSearch
```

src/reducers/filterReducer.js
yingzhao-1: this one actually doesn't work. For some reason, the defaultFilter is not undefined, there is a null value passed into it. So the default parameter will not be assigned to defaultFilter.

test/reducers/filterReducerTest.js
```diff
@@ -33,24 +34,41 @@ const changedEventFilter = { incidentId: 1, ticketId: 0, eventTypes: [1, 2, 3] }
 
 describe('filterReducer', function () {
```
pdimitratos: In general, we should not test combined reducers; we should test each reducer individually. This makes the tests more focused and easy to update as the parts of the combined reducer are updated over time. For most of these reducers, you'd only need to test 3 cases:

* When CHANGE_EVENT_FILTER has a filter value, and the value is different from the state, return the filter value
* When CHANGE_EVENT_FILTER has a filter value, and the value is the same as the state, return the state
* When another action is received (I use { type:'IGNOREME' }), return the state.

You can write this test set once and copy it several times with minor adjustments to match each of the reducers.

That doesn't necessarily need to happen in this pull request, but it will need to happen eventually, as testing the combined reducers introduces much more complexity than testing the individual reducers, and that complexity grows sharply as we add cases to the individual reducers (especially when multiple parts of the combined reducer respond to the same action).

src/components/uncorrelatedEvents.js
```diff
@@ -18,15 +19,15 @@ class UncorrelatedEvents extends Component {
...
-    dispatch(filterActions.clearFilterIncidentId(newFilter))
+    dispatch(filterActions.clearFilterIncidentId(Object.assign({}, filters, {incidentId: null})))
   }
 
   render () {
     const { events, dispatch, eventTypes, history } = this.props
     return (
       <div >
         <TimeAndDatePicker history={history} />
+        <TextFilter history={history} />
```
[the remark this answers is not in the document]
yingzhao-1: no

src/reducers/filterReducer.js
macrobug: who is calling that method, and what value is it passing into it?

src/components/uncorrelatedEvents.js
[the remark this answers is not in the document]
pdimitratos: Would adding it there be more difficult than just adding the TextFilter component in the Timeline component? We should expect that users will be looking at the single-incident view the vast majority of the time, and that view should remain at least at parity with the features on the all-incidents timeline if not better supported.

src/reducers/filterReducer.js
pdimitratos: IIRC, that would be the FilterService method that extracts a filter object from the URL, which is called by configureStore.js. In reducers/index.js,

```JavaScript
  events: events(filters),
```
in the combineReducers statement can become

```JavaScript
  events: filters ? events(filters) : events(),
```

To get around this, though there are other ways to get the same effect. Not sure what's best.

src/components/uncorrelatedEvents.js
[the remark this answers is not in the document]
yingzhao-1: I tried, it could be added onto the timeline easily, the only thing need to be done for this is to update the filter in url.

src/reducers/filterReducer.js
yingzhao-1: so I introduced the initalDefaultFilter, but it broke the filterReducerTest because in the test it's testing filterReducer as a whole. So, should I just rewrite the test to test each reducer in filterReducer?

src/reducers/filterReducer.js
```diff
@@ -1,15 +1,86 @@
...
+const initalFilter = {
+  ticketId: null,
+  incidentId: null,
+  eventTypes: null,
+  startTime: null,
+  endTime: null,
+  DataSearch: null
+}
+
+export const filter = (defaultFilter = initalFilter) => {
```
macrobug: typo: `initalFilter `should be `initialFilter`

src/components/elements/TimePicker.js
```diff
@@ -70,21 +69,24 @@ class TimeAndDatePicker extends Component {
...
-    const { dispatch, filters, history } = this.props
+  setStartAndEndTime () {
+    const { dispatch } = this.props
     const { startDate, startTime, endDate, endTime } = this.state
     let start = (startDate && startTime) ? startDate + 'T' + startTime : null
     let end = (endDate && endTime) ? endDate + 'T' + endTime : null
-    const newFilter = Object.assign({}, filters, { startTime: start, endTime: end })
-    dispatch(filterActions.synchronizeFilters(newFilter, null, null, history))
+    if (!(start && end)) {
+      end = DateTime.local().toISO()
```
macrobug: I see this is using local time. Should this be UTC instead?

src/components/elements/TimePicker.js
```diff
@@ -70,21 +69,24 @@ class TimeAndDatePicker extends Component {
           }}
                 />
         <FlatButton
-          label='submit'
-          primary
-          onClick={() => this.loadUncorrelatedEvents()}
+          label='Select time range'
```
macrobug: Can the user enter date and time manually, without the time pickers?

yingzhao-1: it seems that material-ui does not support the typing in

src/components/elements/TimePicker.js
pdimitratos: We should use UTC everywhere but actual display of time.

PR #110, opened 2018-03-27
[UX] Responsive Layout: make better use of available screen real estate.

macrobug:
**Before:**
image

**After:**
image

### DIFF
--- a/src/components/Search/CreateIncident.js
+++ b/src/components/Search/CreateIncident.js
@@ -12,7 +12,7 @@ const onSubmit = (input, history) => () => {
 }
 
 export const CreateIncident = ({input, creationError, history, dispatch}) => {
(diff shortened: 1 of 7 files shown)


macrobug:
**Before:**
image

**After:**
image

src/components/Search/CreateIncident.js
```diff
@@ -12,7 +12,7 @@ const onSubmit = (input, history) => () => {
 }
 
 export const CreateIncident = ({input, creationError, history, dispatch}) => {
-  return <form id='incident-search' onSubmit={onSubmit(input, history)}>
+  return <form id='incident-search' onSubmit={onSubmit(input, history)} style={{padding: '16px'}}>
```
macrobug: This page is going to be replaced with a "Go To Ticked ID" field eventually.

PR #118, opened 2018-04-02
Remove expand and collapse feature & Ticket compare feature.

macrobug:


src/components/Timeline/AddEventCard.js
```diff
@@ -1,15 +1,12 @@
 import React from 'react'
 import { Card, CardHeader, CardText } from 'material-ui/Card'
+
 import AddEvent from 'components/Timeline/AddEvent'
 
 export const AddEventCard = (incidentId) => <Card>
-  <CardHeader
-    title={'Add Event'}
-    actAsExpander
```
pdimitratos: This is a functional change, let's not mix it in with removing unused code.

src/components/Timeline/Event.js
```diff
@@ -52,15 +52,13 @@ export const Event = ({
...
                   subtitle={timeFormattedToMultipleZones(time)}
-                  actAsExpander
-                  showExpandableButton
                   iconStyle={{
                     color: isAllPlaybookInfoAvailable ? 'black' : 'Lightgrey'
                   }}
             />
                 {
               isAllPlaybookInfoAvailable &&
-              <CardText expandable>
```
pdimitratos: This is a functional change, let's not mix it in with removing unused code.

macrobug: The main goal of this PR was to remove the expand/collapse functionality. The ticket compare removal was just piggy backed here. I can split the PR in two if needed.

PR #119, opened 2018-04-02
Bugfix/signalrfilter

pdimitratos:
Attempted to fix SignalR filter synchronization bug. In the process, converged implementation of event timeline and filter-related components to minimize necessary hooks for signalR updates.


src/actions/eventActions.js
```diff
@@ -14,24 +14,24 @@ export const POST_EVENT_TRY = 'POST_EVENT_TRY'
...
 )
 
 export const fetchEvents = (filter) => reduxBackedPromise(
-        getEventsFetchArgs(filter),
-        getEventsActionSet(filter.incidentId)
-    )
-
-export const fetchUncorrelatedEvents = (filter) => reduxBackedPromise(
-        ['events/' + filterService.serializeFiltersForUrl(filter)],
-        getEventsActionSet(null)
```
yingzhao-1: Did you merged this step into fetchEvents?

pdimitratos: Yes, getEventsFetchArgs will call getEventsEndPoint to generate the base url (/incidents/X/events/ when incidentId has value X or /events/ when incidentId has no value)

src/actions/filterActions.js
```diff
@@ -1,88 +1,102 @@
...
+  dispatch(applyFilter(history, newFilter, signalRFilterType))
 }
 
-export const synchronizeFilters = (filter, incidentId, ticketId, history) => {
-  const newFilter = Object.assign({}, filter, { incidentId: incidentId, ticketId: ticketId })
-  return applyFilter(history)(filter, newFilter)
+export const applyFilter = (history, filter, signalRFilterType) => (dispatch) => {
+  filterService.updateUrlToMatchFilter(history, filter)
+  dispatch(signalRActions.updateEventFilterPreference(signalRFilterType, filter))
+  dispatch(eventActions.clearEvents())
```
yingzhao-1: Why do you clear the current events before fetching new events?

pdimitratos: Because that clears out events that don't match the newly applied filter.

yingzhao-1: We were not clearing out old events previously

src/components/Timeline/EventFooter.js
```diff
@@ -32,4 +34,8 @@ const EventFooter = ({pagination, dispatch}) => {
   )
 }
 
-export default EventFooter
+export const mapStateToProps = (state, ownProps) => ({
+  pagination: state.events.pages
+})
+
+export default connect(mapStateToProps)(EventFooter)
```
yingzhao-1: why do you add mapStateToProps to EventFooter? Where does the pagination value come from previously?

pdimitratos: It was previously passed down from Timeline, but we should take things out of state as close to where they're being used as possible rather than passing them around after taking them out of state. Doing so avoids unnecessary re-rendering within React.

src/actions/filterActions.js
pdimitratos: Previously, CHANGE_EVENT_FILTER cleared out events. Check the EventReducer. The effect of that was that whenever the filter changed, the event collection was dropped.

yingzhao-1:
Code relating to my filters looks good.

src/reducers/filterReducer.js
```diff
@@ -14,47 +16,43 @@ export const incidentId = (defaultIncidentId = initialFilter.incidentId) => (sta
...
-      if (action.startTime) {
-        return action.startTime
-      } else return state
-    default:
-      return state
-  }
-}
+export const startTime = (defaultStartTime) => dateTimeReducer(
+  defaultStartTime ? defaultStartTime.split('T')[1] : DateTime.utc().toISOTime().toString(),
+  defaultStartTime ? defaultStartTime.split('T')[0] : DateTime.utc().minus({day: 1}).toISODate().toString(),
```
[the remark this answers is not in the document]
RedSquirrelious: Here's an area where the variable naming gets a bit confusing.  The other variables names are great, though.  Very clear.

src/components/Timeline/EventFilter.js
```diff
@@ -1,11 +1,14 @@
 import React from 'react'
 import { connect } from 'react-redux'
 import { withRouter } from 'react-router-dom'
-import IconButtonStyled from 'components/elements/IconButtonStyled'
-import FilterChips from 'components/elements/FilterChips'
 import ArrowDown from 'material-ui/svg-icons/navigation/arrow-downward'
 import ArrowUp from 'material-ui/svg-icons/navigation/arrow-upward'
 
+import IconButtonStyled from 'components/elements/IconButtonStyled'
```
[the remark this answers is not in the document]
pdimitratos: Yeah, trying to separate 3rd party from internal dependencies, though I don't think we've codified it.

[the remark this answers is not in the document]
macrobug: I think it is:

```
import third-party
import third-party
import third-party

import our-own
import our-own
import our-own
```

src/reducers/filterReducer.js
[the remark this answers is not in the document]
pdimitratos: That makes sense and should happen eventually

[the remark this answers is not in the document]
RedSquirrelious: In some other PR.  You've got approvals.  Run with it!  :)

strand:
I think this commit may have introduced a regression. When I load `/tickets/:ticketid` I expect it to load the events associated with that ticket. However, after this SHA the URL is changed to something like `/tickets/51948144?startTime=2018-04-08T23:37:00.675Z&endTime=2018-04-10T23:37:00.682Z`, filtering for the past 48 hours of events. Is this supposed to default to filtering to the last two days? 

My assumption is when the URL has no query parameters, that no filters will load by default. Is this correct?

cc @yingzhao-1

PR #120, opened 2018-04-02
Fix yellow highlight

macrobug:
Problem was introduced at commit 17d6718 (Switch from moment.js to Luxon).

### DIFF
--- a/src/components/Timeline/Event.js
+++ b/src/components/Timeline/Event.js
@@ -27,7 +27,7 @@ export const Event = ({
   const itemHighlight = (event && event.timeReceived) ? {
     animationName: 'yellowfade',
     animationDuration: '30s',
-    animationDelay: -(DateTime.utc().diff(event.timeReceived, 'seconds')) + 's'
+    animationDelay: event.timeReceived.diffNow('seconds').toObject().seconds + 's'
   } : {}
   const isAllPlaybookInfoAvailable = !!(actions && Array.isArray(actions) && actions.length > 0)
   const missingId = eventTypeId


test/components/Timeline/EventTest.js
```diff
@@ -55,6 +55,18 @@ describe('Event', function test () {
...
   })
 
+  describe('#animationDelayAsSecondsString', () => {
+    it('returns the appropriate number of seconds', () => {
+      const mockEvent = {
+        timeReceived: DateTime.utc().minus(Duration.fromObject({ seconds: 9 }))
+      }
+
+      const actualString = animationDelayAsSecondsString(mockEvent)
+      expect(parseInt(actualString.slice(0, -1))).to.be.within(-12, -9)
```
strand: 😕 Why was this change necessary? This function shouldn't take three seconds to run, can we restrict the range of acceptable calculated times to be between -9.01s and -9s ago?

`parseInt` will coerce a string beginning with a number to that number without regard to what it ends with, so `slice` is not needed and decreases readability.

macrobug: Regarding

```
expect(parseInt(actualString)).to.be.closeTo(-9, 0.001)
```

`parseInt` will truncate the decimal values, so 9.001 and 9.999 will always be 9. So the `closeTo 0.001` isn't doing much.

Regarding the time, yes, I can constraint it more, will decrease to 2 seconds. But keep in mind this is not a performance test, so it makes sense to allow some room for latency.

Regarding the `slice`, I prefer to not send the 's' character to the parseInt, so I'll keep that there.

Regarding

```
expect(actualString.endsWith('s')).to.be
```

`endsWith` does look more readable, but seems to be missing the `true` at the end. Will make that change and push.

strand: > `parseInt` will truncate the decimal values, so 9.001 and 9.999 will always be 9. So the closeTo 0.001 isn't doing much.

>_<

strand:
Hey, I wanted to let you know _why_ I used the `.to.be` syntax. In RSpec 2, `.to be` is equivalent to `.to be(true)`. I think this was not well understood or used, as it was removed in RSpec 3, and but also RSpec and Chai are close but not equivelant. Specifically, the `be` function in Chai is _just_ used for chaining and does not check for existence. 

Anyhow, just wanted to say thanks for shining some light on a past understanding that had gotten stale and wrong. I'll remember in the future that be in Chai is not a matcher, it's just chaining.

PR #122, opened 2018-04-05
Fix: when filter is applied via URL, the app won't load due to initial values undefined.

macrobug:
image

### DIFF
--- a/src/reducers/filterReducer.js
+++ b/src/reducers/filterReducer.js
@@ -1,7 +1,16 @@
 import * as filterActions from 'actions/filterActions'
 import { combineReducers } from 'redux'
 
-export const incidentId = (defaultIncidentId) => (state = defaultIncidentId, action) => {
+const initialFilter = {
+  ticketId: null,
+  incidentId: null,
+  eventTypes: null,
+  startTime: null,
+  endTime: null,
+  dataSearch: null
+}
+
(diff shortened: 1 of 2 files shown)


src/reducers/filterReducer.js
```diff
@@ -69,23 +78,14 @@ export const ticketId = (defaultTicketId) => (state = defaultTicketId, action) =
...
-const initialFilter = {
-  ticketId: null,
-  incidentId: null,
-  eventTypes: null,
-  startTime: null,
-  endTime: null,
-  DataSearch: null
-}
-
 export const filter = (defaultFilter = initialFilter) => {
```
yingzhao-1: Since initialFilter value is assigned to child reducers, it need not to be assigned to defaultFilter here.

test/reducers/filterReducerTest.js
```diff
@@ -4,7 +4,47 @@ import { expect } from 'chai'
...
+      const defaultFilter = {
+        ticketId: null,
+        incidentId: null,
+        eventTypes: null,
+        startTime: null,
+        endTime: null,
+        dataSearch: null
+      }
+
+      const result = filterReducer.filter(defaultFilter)({}, { type: null, filter: null })
```
yingzhao-1: wasn't it work like this previously?

src/reducers/filterReducer.js
macrobug: I see what you mean. I'll still leave it there because it covers the case where defaultFilter is undefined.

PR #127, opened 2018-04-09
Readied for build wrapper

RedSquirrelious:
Addition of distConstants.js works with build script.  Changes to base.js and dist.js needed for build wrapper.  We weren't using Karma et al. and it was causing bloat in build.


cfg/base.js
```diff
@@ -31,7 +31,7 @@ const config = {
     publicPath: publicPath
   },
   devServer: {
-    contentBase: './src/',
+    contentBase: path.join(siaRoot, 'src'),
```
strand: Nice!

cfg/distConstants.js
```diff
@@ -0,0 +1,14 @@
...
+  appEnv: 'dist',
+  aadInstance: argv.aadInstance,
+  aadTenant: argv.aadTenant,
+  clientId: argv.clientId,
+  baseUrl: argv.baseUrl,
+  authRedirectUri: argv.authRedirectUri,
+  authVersion: argv.authVersion
+})
+
+module.exports = distConstants
```
strand: Add a newline at end-of-file.

package.json
```diff
@@ -90,15 +90,6 @@
     "eslint": "^4.18.2",
     "eslint-plugin-react": "^7.7.0",
     "file-loader": "^1.1.6",
-    "karma": "^2.0.0",
```
strand: Why are these dependancies necessary?

RedSquirrelious: We were using/planning to use Karma but ended up running our tests straight from Mocha.  These aren't needed right now and were slowing down builds in the new system.  They can always be re-added in the future if we use Karma.

strand: Nevermind. I'm cutting back on my caffeine and I thought red was green.

PR #129, opened 2018-04-09
Display ticket severity

danieljurek:
**Questions**
* Do we need to be concerned about the scenario where `ticket.data` is `null`? This change accounts for that but it may be redundant. 
* Given that this change is small and we're about to redo layout, is this change acceptable without tests at this time? Or is the presence of branching logic sufficient to justify a test? 

**Summary**
* We previously tried to display severity but were accessing it at the wrong place. This fixes the issue by accessing the severity in the `ticket.data` object and handles situations where the ticket doesn't have a `severity` property. 
* I spoke with Alberto about color coding... that is not in scope for this item

**What it looks like**
image


pdimitratos:
Yes, null data can happen when the ticketing connector fails or when there's no ticket with the ID we're looking for.

We should be fine to merge this without tests because there are no tests for IncidentSummary yet. Writing those tests would be useful, but doesn't need to be part of this pull.

PR #130, opened 2018-04-10
Fix: Sia Buttons activate when text is clicked but not when button is clicked

macrobug:
**Before:**
image

**After:**
image


macrobug:
Fixing the broken tests....

PR #131, opened 2018-04-10
GoToTicket box is shown on all pages

pdimitratos:
For Task 47549: Replace "Incident Search" page with "Go to Ticket" field.

This is extremely unlikely to be the style/layout we want, but is this the correct functionality? GoToTicket is shown on all pages.


macrobug:
For some reason this doesn't run on my machine:

image

macrobug:
Managed to build and run on my local. This seems to be the correct functionality.

Regarding layout, something like this probably works:
image

Also, make sure to remove the search page, and remove it from the menu as well.

PR #132, opened 2018-04-11
Show incident title in menu

danieljurek:


src/components/elements/MenuLink.js
```diff
@@ -4,18 +4,24 @@ import { Link } from 'react-router-dom'
...
-      primaryText={`${type} ${id}`}
+      primaryText={ticketPrimaryText(type, id, title)}
       rightIcon={<ActionDelete onClick={() => dispatch(onClick(id))} />}
     />
   )
 }
 
+const ticketPrimaryText = (type, id, title) => (
+  <div className='issue-title-text'>
+    {`${type}-${id}: ${title}`}
```
macrobug: I don't think the type needs to be shown to the user. What other types apart from tickets do we expect to display here?

src/styles/App.css
```diff
@@ -37,6 +37,12 @@ div#siaApp {
   display: none;
 }
 
+.issue-title-text {
```
macrobug: The preferred styling approach in Material-UI is `CSS-in-JS`. 

Unless there is a compelling reason to put the style in App.css (e.g. the style is global instead of just a local thing), then we should try to keep the style as close as possible to its target element, and keep the App.css as minimal as possible.

See here for more info: 
https://material-ui-next.com/customization/css-in-js/
https://medium.com/seek-blog/a-unified-styling-language-d0c208de2660
https://medium.freecodecamp.org/css-in-javascript-the-future-of-component-based-styling-70b161a79a32

src/components/elements/MenuLink.js
```diff
@@ -4,18 +4,21 @@ import { Link } from 'react-router-dom'
...
       key={`${type}-${id}`}
       containerElement={<Link to={typeRoute} />}
-      primaryText={`${type} ${id}`}
+      primaryText={ticketPrimaryText(type, id, title)}
       rightIcon={<ActionDelete onClick={() => dispatch(onClick(id))} />}
     />
   )
 }
 
+const ticketPrimaryText = (type, id, title) => `${type} ${id}: ${title}`
```
macrobug: Probably `type` is not needed. It just takes up space, and it's probably going to be `ticket` always anyways.

danieljurek:
Ready for a real review

A few things to call out: 
* **We might be testing the wrong things** -- That is, a lot of our tests are coupled directly to the ordering of lists of components, components instantiated by tested components (coupling tests for component A to the implementation of component B), and defined props on components. It might make sense to focus testing efforts toward business logic (like that in our `mapStateToProps` functions and functions that have branching logic like [`ticketMenuLinks`](https://github.com/Azure/Sia-EventUI/pull/132/files#diff-d083044947dca9e192eaf81e4548458aR41). There are ways to automate testing that components render to the DOM in a certain way with given inputs that doesn't require comparing images. 
* The `MenuLink` component is expected to do a lot... I think @pdimitratos and I were able to get it defined well enough 
* The `MenuItem.primaryText` in [`MenuLink`](https://github.com/Azure/Sia-EventUI/pull/132/files#diff-b8bb757780e4d5476ffd77942ee4f1beR19) is treated as a `node` instead of plain text. This is necessary in order to achieve the desired effect. The `innerDivStyle` property of `MenuItem` is not sufficient: the `innerDivStyle` div includes the icon and the text formatting CSS doesn't propagate down to the div that contains the `primaryText`... The `primaryText` item should have a name that is more descriptive of its flexability

PR #134, opened 2018-04-12
Refactored event display components and added event icons

pdimitratos:
This now has tests and should be ready for review


src/components/Timeline/Event/Events.js
```diff
@@ -0,0 +1,77 @@
+import React from 'react'
```
strand: I'm not sure why GitHub isn't picking up that this was a file moved from the `Timeline` folder. This is the evolution of `src/components/Timeline/Events.js`, jäh?

src/components/Timeline/Event/Events.js
```diff
@@ -0,0 +1,77 @@
...
+
+import ErrorMessage from 'components/elements/ErrorMessage'
+import LoadingMessage from 'components/elements/LoadingMessage'
+import Highlight from 'components/elements/Highlight'
+import EventCard from 'components/Timeline/Event/EventCard'
+import BootstrapPlaybook from 'components/Timeline/Playbook/BootstrapPlaybook'
+import * as eventTypeActions from 'actions/eventTypeActions'
+
+export const Events = ({events, ticketId, incidentId}) => <div>
+  {Array.from(events)
```
strand: Are we expecting `events` to have a type other than Array? 

Would the code below give the same results?
``` javascript
  {
    const connectEvent = (event) => <ConnectedEvent 
      key={event.id} 
      event={event} 
      ticketId={ticketId} 
    />

    Array.from(events).map(event => connectEvent(event))
  }
```

src/components/Timeline/Event/Events.js
```diff
@@ -0,0 +1,77 @@
...
+
+const eventHasValidDisplayText = (event) =>
+  event && event.data && event.data.DisplayText
+
+export const Event = ({
+  eventTypeIsFetching,
+  eventTypeIsError,
+  event,
+  ticketId
+}) => eventTypeIsFetching && !eventHasValidDisplayText(event)
```
strand: Move the negation into `eventHasValidDisplayText` and rename to `eventHasInvalidDisplayText`.

test/components/Timeline/Event/EventsTest.js
```diff
@@ -0,0 +1,149 @@
...
+      events: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10].map(id => ({id})),
+      ticketId: 2,
+      incidentId: null
+    }
+
+    const result = createComponent(Events, input)
+
+    it('Should return a div containing an array of Events', function () {
+      expect(result.type).to.equal('div')
+      expect(result.props.children[0].type).to.equal(ConnectedEvent)
```
strand: Is there a way we can test this without exporting a ConnectedEvent? Is it valuable to test? [⎁](tktktk)

src/components/Timeline/Event/Events.js
```diff
@@ -0,0 +1,77 @@
...
+
+  return {
+    eventTypeIsFetching: state.eventTypes.fetching.includes(event.eventTypeId),
+    eventTypeIsError: state.eventTypes.error.includes(event.eventTypeId),
+    event,
+    ticketId
+  }
+}
+
+export const ConnectedEvent = connect(mapStateToEventProps)(Event)
```
strand: [⎂](tktktk) This could be inlined to the scope which it is used in. About 50 lines above it.

src/components/Timeline/Event/Events.js
pdimitratos: Yes

macrobug: The way git tracks renames is by looking at file content. If the file was both renamed and substantially modified, then it has no way of knowing the previous one was an ancestor.

src/components/Timeline/Event/EventCard.js
```diff
@@ -0,0 +1,62 @@
...
+>
+  <CardHeader
+    title={titleText}
+    subtitle={timeFormattedToMultipleZones(DateTime.fromISO(event.occurred ? event.occurred : event.Occurred))}
+    iconStyle={{
+      color: 'black'
+    }}
+    avatar={IconType ? <Avatar icon={<IconType />} /> : null}
+  />
+  <CardText>
```
macrobug: This could be CardActions instead of CardText. This keeps it semantic, and also the padding is more compact.

src/components/Timeline/Event/Events.js
pdimitratos: Events should always be an array.

That code would give the same result. Do you think that improves readability? Are there other benefits from that format? I'm somewhat partial to inlining the components in map statements within JSX.

src/components/Timeline/Event/Events.js
pdimitratos: I will be more careful about this in future

test/components/Timeline/Event/EventsTest.js
pdimitratos: It's not possible to test it accurately without exporting the component (ConnectedEvent) to the best of my knowledge. It is valuable to test because I have often accidentally made changes in such a way that I was using the non-connected version of a component where I was intending to use the connected version; this test would catch that.

src/components/Timeline/Event/Events.js
pdimitratos: Rearranged this file so that Events comes immediately after ConnectedEvent

test/components/elements/ErrorMessageTest.js
```diff
@@ -1,75 +1,75 @@
...
 import { Card, CardHeader } from 'material-ui/Card'
 import ErrorIcon from 'material-ui/svg-icons/alert/error'
 import { RetryButton } from 'components/elements/Buttons'
 
 describe('ErrorMessage', function () {
-  context('when inputs are valid', function() {
+  context('when inputs are valid', function () {
     describe('when given only a message', function () {
-      const testObject = ErrorMessage('TestMessage')
+      const testObject = createComponent(ErrorMessage, {message: 'TestMessage'})
```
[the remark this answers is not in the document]
pdimitratos: Do you mean renaming the test constant or renaming something within ErrorMessage?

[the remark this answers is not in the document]
strand: `testObject` could be renamed to `errorMessage` as the object under test.

src/components/Timeline/Event/Events.js
macrobug: @strand, if more than 50% of a file's content has been modified, then git won't track the rename. The use of `git mv` won't make a difference. The recommended approach is to do the rename in one commit, and the modification in another one.

If you are still curious, this is an easy thing to test:

```
cd Sia-EventUI
git mv README.md README2.md
git status
   renamed:    README.md -> README2.md
```

Now open the file and make some changes. Then:

```
git add .
git status
```

If the file was slightly modified it should still say:

```
   renamed:    README.md -> README2.md
```

If the file was substantially modified, it should say:

```
   deleted:    README.md
   new file:   README2.md
```

The reason this happens is because git is a content tracker, not a file tracker, so it doesn't really care about file or directory names.

The default threshold is 50%, but it is possible to list renames based on a different threshold, for example 30%:
```
git log -M30 -p --stat
```

PR #135, opened 2018-04-13
Style Playbook Actions

macrobug:
**Before:**
image

**After:**
image


src/components/Timeline/Playbook/DisplayPlaybook.js
```diff
@@ -11,53 +14,38 @@ export const DisplayPlaybook = ({
...
-  }
-}
+export const DisplayAction = (
+  action,
+  eventTypeId,
+  eventId,
+  ticketId,
+  incidentId,
+  key
+) => <div key={key}>
```
pdimitratos: Is this div needed? Could we pass the key to the Play component instead?

PR #136, opened 2018-04-13
added incident status and adjusted summary format

yingzhao-1:


src/components/Incident/IncidentSummary.js
```diff
@@ -42,11 +42,22 @@ const BasicInfoColumn = (ticketSystem, ticket) => [
...
+    ticket.status ? <font key={key}> | </font>
+      : null,
+  (key) => {
+    if (ticket.status) {
+      const fontColor = ticket.status === 'Active' ? 'red' : 'black'
+      return <font key={key} color={fontColor}>
+        {ticket.status}
+      </font>
+    } else return null
+  }
```
danieljurek: A few things to make this code more HTML compliant and easier to read: 
* The `<font>` tag is no longer supported https://www.w3schools.com/tags/tag_font.asp, we should use CSS. And our CSS usage belongs inside the component (the `styles` object in the code sample below) 
* The method for outputting the pipe `|` character can be folded into the ticket status method (just like you did above with the severity which avoids having two functions that check the same `ticket.status` value and collapses them into one function) 
* The font coloring and ticket status function has 3 different paths and a conditional nested within a conditional. We can simplify this logic by putting the ticket status highlighting in its own function, and now each function has only one layer of condition checking.

Another way to express the behavior could be:  
```
// Below the `import` statements, define the styles
const styles = {
    activeStatus: { color: 'red' }
}

... 
// inside of BasicInfoColumn where we show the ticket status, use a function to reduce the number of conditionals the method has to evaluate 
  (key) => ticket.status 
      ? highlightedTicketStatus(key, status) 
      : null 

... 
// Outside of the `BasicInfoColumn` function: Fold the pipe character `|` displaying inside the ticket status display method
const highlightedTicketStatus = (key, status) => (
    <div key={key}>
        | <span style={status === 'Active' ? styles.activeStatus : null}>{status}</span>
    </div>
)
```

PR #138, opened 2018-04-13
[WIP] Add checklist

johana-star:
This work-in-progress needs to meet several criteria before it is ready for review:

- [ ] extract anonymous functions into named variables.
- [ ] The current implementation is global, change so that checklists are tracked per-ticket.
- [ ] Why did past Strand not test this? 😢Please add tests to cover the lifecycle of the Checklist toggle at least.
- [ ] Link to 'binding the callback' in React documentation
- [ ] Do a quick review for clarity.
- [ ] @macrobug has noted approaches which remove the lodash dep. Remove it.
- [ ] Also, update the README to indicate that we prefer plain-old EcmaScript to using lodash or another library for functional transformations.
- [ ] Look at using a solution other than `localstorage`—specifically look for a drop-in replacement which will encrypt user data so other sites won't be able to read the task list stored in the user's browser.
- [x] Inline styles are not cool. replace inline styles with class names.

### DIFF
--- a/src/components/Checklist.js
+++ b/src/components/Checklist.js
@@ -1,13 +1,14 @@
 import React from 'react'
 import _ from 'lodash'
 
+require('../styles/Checklist.css')
+
 // Stateless, functional components. It is their responsibility to tranform
 // data for display. They should be designed so they will not raise an error,
 // and they must be contained by ErrorBoundaries if they might raise.
-const Header = ({title}) => <h2>{title}</h2>
 const className = (item) => item.checked ? 'item checked' : 'item unchecked'
 
-const Title = (ticketId) => <span>
+const Title = (ticketId) => <h2>
   Checklist {ticketId ? ` for ${ticketId}` : ''}
   <a
     style={{ margin: '0 0 0 10px' }}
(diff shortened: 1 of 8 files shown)


src/styles/App.css
```diff
@@ -14,11 +14,21 @@ div#siaApp {
   background: white;
 }
 
+.item {
```
alrodrig1: The preferred styling approach in Material-UI is `CSS-in-JS`. 

Unless there is a compelling reason to put the style in App.css (e.g. the style is global instead of just a local thing), then we should try to keep the style as close as possible to its target element, and keep the App.css as minimal as possible.

See here for more info: 
https://material-ui-next.com/customization/css-in-js/
https://medium.com/seek-blog/a-unified-styling-language-d0c208de2660
https://medium.freecodecamp.org/css-in-javascript-the-future-of-component-based-styling-70b161a79a32

src/components/Checklist.js
```diff
@@ -0,0 +1,80 @@
...
+    this.toggleCheck = this.toggleCheck.bind(this);
+  }
+
+  toggleCheck = (ev) => {
+    const isChecked = () => ev.currentTarget.classList.contains('checked')
+    const content   = () => ev.currentTarget.textContent
+    this.itemToUpdate = { text: content(), checked: !isChecked() }
+    ev.preventDefault()
+    this.setState((pastState) => {
+      let index = _.findIndex(
```
alrodrig1: No need to add lodash here. Regular ES works just fine and IMO is more clear:
```
pastState.items.map(item => item.text).indexOf(this.itemToUpdate.text)
```

src/components/Checklist.js
```diff
@@ -0,0 +1,80 @@
...
+      return pastState
+    })
+  }
+
+  render() {
+    return <div id='checklist'>
+      <Header title={Title()} />
+      <ul>
+        {
+          _.chain(this.state && this.state.items).
```
alrodrig1: I'm not sure why you need lodash or `_.chain` here. Regular ES works just fine, is shorter, and IMO is more clear:

```
this.state.items.map((item, index) =>
  <li
    style={{listStyle: "none"}}
    className={className(item)}
    onClick={this.toggleCheck}
    key={ index }>{ item.text }
  </li>)
```

src/components/Checklist.js
```diff
@@ -0,0 +1,80 @@
+import React from 'react'
+import _ from 'lodash'
```
alrodrig1, line 2: This is being used here, but is not in the package.json file. However, before you add it, see comments below.

alrodrig1 (changes requested):

johana-star:
This work is cancelled.

over

PR #144, opened 2018-04-18
Add snapshot testing for MenuLink

danieljurek:
Small example of using Snapshot Testing to eliminate bespoke brittle brittle tests. 

Mocha seems to include global level configs with just one call to `import 'test/helpers/enableSnapshot'` in any test.js file (test execution order doesn't matter), however I recommend including the call to `import` enableSnapshot in all files that use snapshot testing. Mocha is smart enough to not double up on calling functions registered with `beforeEach` if they've been registered before.


test/helpers/enableSnapshot.js
```diff
@@ -0,0 +1,12 @@
+import chai from 'chai'
+import chaiJestSnapshot from 'chai-jest-snapshot'
+
+chai.use(chaiJestSnapshot)
+
+before(function () {
```
strand: I dislike hiding any magic with test helpers, and before hooks definitely fall into that territory for me.

However, this is a pretty good extraction, and I don't see any reason not to keep it. I think we should rename files, so it is clear from far away where the magic happens. This file might be `enableSnapshotUsingBeforeHooks.js` while `test/components/elements/MenuLinkTest.js` becomes `test/components/elements/MenuLinkSnapshotTest.js`. I'm not sure tho… what do you think?

test/components/elements/MenuLinkTest.js
```diff
@@ -1,47 +1,31 @@
 'use strict'
+import React from 'react'
 import { expect } from 'chai'
-import { Link } from 'react-router-dom'
-import MenuItem from 'material-ui/MenuItem'
 import MenuLink from 'components/elements/MenuLink'
+import { shallow } from 'enzyme'
+import toJson from 'enzyme-to-json'
+require('test/helpers/enableSnapshot')
```
strand: Or maybe, we `import { setupSnaphot, resetSnapshotConfigBeforeEach } from 'test/helpers/enableSnapshot'`…

Then we can call setupSnapshot here, and resetSnapshotConfigBeforeEach within the context of the decribe block where the snapshot happens… 🤔

test/helpers/enableSnapshot.js
danieljurek: Renaming to `enableSnapshotUsingBeforeHook.js`: agreed 
Renaming snapshot tests... need more information: Business logic tests can also live in that file because we're testing the whole component... in this case there's no need for business logic tests on that component, but something more complex like `NavMenu` would still need business logic tests for `mapStateToProps`, etc.

danieljurek:
@strand totally, we should prevent helper abuse. How about using `snapshotScope` to scope calls to hooks so the only requirement is to wrap your tests in a `snapshot` scoping and the hooks aren't getting called on every test. 

...maybe use a name that isn't `snapshotScope`

